PR #17, opened 2016-07-27
Update explainer.md

bsittler:
Changes I inadvertently neglected in https://github.com/bsittler/async-cookies-api/pull/16 :
 - Fix bit-rot in countMatchingSimpleOriginCookies example
 - Add missing url+getAll examples (countMatchingCookiesForRequestUrl and countAllCookiesForRequestUrl)

PR #18, opened 2016-07-27
matchType uses 'startsWith', not 'prefix'

inexorabletash:
Probably a leftover from a previous syntax.


inexorabletash:
Oh, you have this already in #17. Never mind.

bsittler:
Yeah, I'll go ahead and merge that to reduce confusion

PR #25, opened 2016-08-09
Added non-ASCII examples; clarify path handling

bsittler:
While working on https://github.com/bsittler/async-cookies-api/pull/24 I noticed some badly worded parts of the explainer and the lack of non-ASCII examples

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -55,9 +55,9 @@ The API must interoperate well enough with existing cookie APIs (HTTP-level, HTM
 
 ### Opinions
 
-This API defaults cookie paths to '/' for write and delete operations. The implicit relative path-scoping of cookies
-to '.' has caused a lot of additional complexity for relatively little gain given their security equivalence under the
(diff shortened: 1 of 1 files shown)


bsittler:
@domenic @annevk What do you think of these proposed changes? I believe this clarifies path handling for read and write operations and gives examples with non-ASCII cookie values which should better explain the intent

domenic:
I guess I'm not sure I entirely understand the processing model for the path stuff, but I can wait for precise spec text for that.

Non-ASCII examples are good. Maybe you want to give examples of how writing allows you to read back and get the same values?

bsittler:
> I guess I'm not sure I entirely understand the processing model for the path stuff, but I can wait for precise spec text for that.

Yeah, this part needs a spec badly. Existing browser behavior appears to be somewhat inconsistent here.

> Non-ASCII examples are good. Maybe you want to give examples of how writing allows you to read back and get the same values?

Done, I think. Does it do what you meant?

domenic:
Yeah, looks good!

PR #40, opened 2017-05-08
Service worker and browser cycle in the presence of monitored cookie expiration

bsittler:
This change attempts to clarify expected behavior when monitored cookies expire.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -593,7 +593,7 @@ addEventListener('cookiechange', event => {
 
 The power and resource consumption of this feature of the API will need to be carefully balanced against its utility, and for this reason delivery of a change event may be deferred until after the next browser start-up (for unrelated reasons) in cases where the change notification is for a cookie expiration and the browser is not already running.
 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -589,6 +591,10 @@ addEventListener('cookiechange', event => {
 
 *Note:* a service worker script needs to be prepared to handle "duplicate" notifications after updates to the service worker script with overlapping cookie change interests compared to those interests previously registered.
 
+The power and resource consumption of this feature of the API will need to be carefully balanced against its utility, and for this reason delivery of a change event may be deferred until after the next browser start-up (for unrelated reasons) in cases where the change notification is for a cookie expiration and the browser is not already running.
+
+This API will cause new service worker start-up after browser start-up in cases where a cookie monitored by the service worker expired while the browser was not running. This will not block other parts of browser start-up, however.
```
mkruisselbrink: presumably the "Expiration of a cookie monitored by a service worker might not deliver the cookie change event until the next time a document or service worker in the cookie's domain or origin is used" comment from further up also still applies here (or in other words, it will not necessarily cause new service worker start-up after browser start-up if nothing else is loaded on that origin during the browsing session)?

bsittler: Correct. Added clarification

PR #47, opened 2017-08-18
Update explainer.md

bsittler:
Cover the rationale for name prefix matching and name matching


bsittler:
Does this explain the rationale adequately, @pwnall  ?

pwnall:
Everything here is useful!

PR #63, opened 2018-06-28
IDL fixes for the specification.

pwnall:
* cookieStore.getAll() is guaranteed to return a CookieList
* CookieListItem's defaults are specified, because they're observable
  when initializing cookie change events (to be spec'd)
* httpOnly doesn't make sense on CookieStoreDeleteOptions

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -216,8 +216,8 @@ interface CookieStore : EventTarget {
   Promise<CookieListItem?> get(USVString name);
   Promise<CookieListItem?> get(optional CookieStoreGetOptions options);
 
-  Promise<CookieList?> getAll(USVString name);
-  Promise<CookieList?> getAll(optional CookieStoreGetOptions options);
+  Promise<CookieList> getAll(USVString name);
+  Promise<CookieList> getAll(optional CookieStoreGetOptions options);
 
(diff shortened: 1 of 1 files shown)


pwnall:
@inexorabletash I'm not confident in my WebIDL-fu yet. Please review?

index.bs
```diff
@@ -271,17 +271,16 @@ dictionary CookieStoreDeleteOptions {
   USVString? domain = null;
   USVString path = "/";
   boolean secure = true;
-  boolean httpOnly = false;
   CookieSameSite sameSite = "strict";
 };
 
 dictionary CookieListItem {
   USVString name;
```
inexorabletash: Good point re: the CookieChangeEvent constructor. 

Should `name`/`value` be required then?

pwnall: Done. Thanks for the catch!

pwnall: I spoke too soon. `value` cannot be required, because it's missing for deleted cookie info. I updated the PR to reflect this.

We could do an empty value instead of missing value key, if you think that'd more usable, but that's a separate discussion.

inexorabletash:
Looks good, but let me run it through bikeshed tomorrow which will do a syntax check too. (And update the actual spec)

PR #64, opened 2018-07-09
Auto-deploy HTML spec

rtoy:
First checkin of basic config and scripts.  Nothing is working yet
because the deploy key hasn't been set up yet.


rtoy:
Abandon this in favor of #65.

PR #86, opened 2018-07-30
Fix #85: compile.sh exits if anything fails

### LINKED ISSUE #85: bikeshed errors should tell Travis that the build failed
Look at https://travis-ci.org/WICG/cookie-store/builds/409965529. There are bikeshed errors which should have stopped the build, but Travis continues and checks in the new html version.  In some ways, this is ok, but we should be notified about bikeshed producing errors.

rtoy:
Also, if the out directory doesn't exist, don't copy anything there.
Mostly useful when running local so we don't try to copy things to the
non-existent directory.

### DIFF
--- a/compile.sh
+++ b/compile.sh
@@ -3,9 +3,15 @@
 # So we can see what we're doing
 set -x
 
-# Run bikeshed.
+# Exist with nonzero exit code if anything fails
+set -e
+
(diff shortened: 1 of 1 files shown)


pwnall:
@rtoy Thank you very much for this fix!

PR #87, opened 2018-08-02
Remove httpOnly option. Fixes #41

### LINKED ISSUE #41: Should a ServiceWorker be able to create HttpOnly cookies?
Related to https://github.com/WICG/async-cookies-api/issues/37

Should this API (working in the world of JavaScript) be allowed to create `httpOnly` cookies?

This is currently not possible from JavaScript within the web page:

    document.cookie = 'session1=abc-123; httpOnly';
    document.cookie = 'session2=abc-123';

    document.cookie
        "session2=abc-123"

It's just a mild concern about [session fixation](https://www.owasp.org/index.php/Session_fixation); as it could be used by someone malicious to create a `httpOnly` session cookie on the victims computer, where the attacker will then know what the value will be (as they set it)... that's not to say that they couldn't set the cookie without the `httpOnly` flag.

inexorabletash:
Per conversation in #41, the `httpOnly` option should be removed. It would allow creation of HttpOnly cookies from the `CookieStore` API, which is something `document.cookie` does not allow.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/87.html" title="Last updated on Aug 2, 2018, 7:13 PM GMT (f856542)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/87/37c1f2c...f856542.html" title="Last updated on Aug 2, 2018, 7:13 PM GMT (f856542)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -345,7 +345,7 @@ await cookieStore.set({
   path: '/',
 
   // Creates secure cookies by default on secure origins.
-  secure: true
+  secure: true,
 });
 ```
 
--- a/index.bs
+++ b/index.bs
@@ -345,8 +345,7 @@ await cookieStore.set({
   path: '/',
 
   // Creates secure cookies by default on secure origins.
-  secure: true,
-  httpOnly: false,
+  secure: true
 });
 ```
 
@@ -568,7 +567,6 @@ dictionary CookieStoreSetOptions {
   USVString? domain = null;
   USVString path = "/";
   boolean secure = true;
-  boolean httpOnly = false;
   CookieSameSite sameSite = "strict";
 };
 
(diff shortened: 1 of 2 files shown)


inexorabletash:
WPT updates in https://chromium-review.googlesource.com/c/chromium/src/+/1160927

This was untested, and actually failed in Chrome's impl (the `set()` call rejected)

index.bs
```diff
@@ -345,8 +345,7 @@ await cookieStore.set({
   path: '/',
 
   // Creates secure cookies by default on secure origins.
-  secure: true,
-  httpOnly: false,
+  secure: true
```
pwnall: nit: Would you be OK with retaining the trailing comma, to minimize future diffs?

(I can imagine that new devs would be a bit confused by this, so it's a trade-off)

inexorabletash: Yeah, can do.

PR #89, opened 2018-09-11
Define Service Worker integration (more) correctly. Fixes #88.

### LINKED ISSUE #88: spec should handle service worker lifetime issues
The cookie store spec is currently a bit vague about what happens when a cookiechange event fires in a service worker context.  It seems to imply that the SW will be started in that case, but does not explicitly spec this.  Any events fired that start the SW should also extend ExtendableEvent so that the handler can keep the worker alive with waitUntil().

This is probably largely solvable by using:

https://w3c.github.io/ServiceWorker/#fire-functional-event-algorithm

And extending for events fired at the SW:

https://w3c.github.io/ServiceWorker/#extendableevent-interface

A bit awkward that these should not have the waitUntil in other contexts, though.

inexorabletash:
* CookieStore.onchange is only in windows
* subscribeToChanges() rejects if url not contained in SW's scope
* Define ExtendableCookieChangeEvent
* Factor out cookie matching
* Enumerate SW registrations, filter cookies by subscriptions,
   and use "fire a functional event" hook.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/89.html" title="Last updated on Sep 12, 2018, 4:14 PM GMT (a8e73e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/89/1362c19...a8e73e9.html" title="Last updated on Sep 12, 2018, 4:14 PM GMT (a8e73e9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -543,6 +543,7 @@ interface CookieStore : EventTarget {
   [Exposed=ServiceWorker]
   Promise<sequence<CookieStoreGetOptions>> getChangeSubscriptions();
 
+  [Exposed=Window]
   attribute EventHandler onchange;
 };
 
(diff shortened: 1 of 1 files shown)


inexorabletash:
@pwnall, @wanderview - can you take a look?

index.bs
```diff
@@ -874,9 +875,12 @@ The <dfn method for=CookieStore>subscribeToChanges(|subscriptions|)</dfn> method
     1. [=list/For each=] |entry| in |subscriptions|, run these steps:
         1. Let |name| be |entry|'s {{CookieStoreGetOptions/name}} member.
         1. Let |url| be |entry|'s {{CookieStoreGetOptions/url}} member.
+        1. If |url| does not start with |registration|'s [=service worker registration/scope url=],
+            then reject |p| with a {{TypeError}} and abort these steps.
```
wanderview: Does this URL get parsed?  Are relative URLs evaluated against a base URL somewhere?  This needs to happen before comparing to the scope string.

inexorabletash: Added a snippet about that. (There's likely a ton of such things missing.)

inexorabletash:
I talked through the changes with @pwnall so I'll go ahead and land, nothing is written in stone.

PR #123, opened 2020-02-05
Explainer fixes

### LINKED ISSUE #118: Explainer: unsubscribe is called in a loop
The explainer has:

```js
 for (const subscription of subscriptions)
    await self.registration.cookies.unsubscribe(subscription);
```

... but in Chrome the IDLs have `unsubscribe()` taking a sequence, so the loop should not be necessary. Which is correct?

### LINKED ISSUE #119: Explainer uses `await` in non-async function
Explainer has 
```js
self.addEventListener('activate', (event) => {
  // Snapshot current state of subscriptions.
  const subscriptions = await self.registration.cookies.getSubscriptions();
  ....
```
This should be `async (event) => ...` for the `await` to be valid.

### LINKED ISSUE #120: Explainer has truncated sentence
Explainer has:

> While the previous samples are mostly aimed at updating a page's UI to reflect

... and it ends there.

pwnall:
1) Fixes call to CookieStoreManager.unsubscribe(), which now receives
   a list of subscriptions.
2) Fixes all instances of await being used in non-async functions.
3) Fixes unfinished sentences noticed along the way.

Fixes #118, Fixes #119, Fixes #120

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/123.html" title="Last updated on Feb 5, 2020, 7:25 PM UTC (cd6de5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/123/f9c7e21...cd6de5f.html" title="Last updated on Feb 5, 2020, 7:25 PM UTC (cd6de5f)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -40,16 +40,12 @@ to session state changes, to clean up private cached data.
 Cookies have also found a niche in storing user decisions to opt out of tracking
 by ad networks, and receive less personalized ads.
 
-Separetly, from a conceptual angle, a service worker is intended to be an
+Separately, from a conceptual angle, a service worker is intended to be an
 HTTP proxy for the pages under its scope. By this principle, service workers
 must be able to read and modify the cookies accessible to pages under their
 scopes.
 
 ### Reacting to session state changes
-
-While the previous samples are mostly aimed at updating a
-page's UI to reflect
-
 
 The following code illustrates synchronous polling via `document.cookie`. The
 code periodically induces jank, as `document.cookie` is a synchronous call
@@ -169,11 +165,11 @@ The following code snippet uses the Cookie Store API instead, and does not jank
 the main thread.
 
 ```javascript
-document.getElementById('opt-out-button').addEventListener('click', () => {
+document.getElementById('opt-out-button').addEventListener('click', async () => {
   await cookieStore.set({ name: 'opt_out', value: '1',
                           expires: new Date('Wed, 1 Jan 2025 00:00:00 GMT') });
 });
-document.getElementById('opt-in-button').addEventListener('click', () => {
+document.getElementById('opt-in-button').addEventListener('click', async () => {
   await cookieStore.delete({ name: 'opt_out' });
 });
 ```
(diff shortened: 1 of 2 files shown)


explainer.md
```diff
@@ -380,19 +378,20 @@ Calls to `subscribe()` and `unsubscribe()` are idempotent.
...
-  for (const subscription of subscriptions)
-    await self.registration.cookies.unsubscribe(subscription);
-
-  await self.registration.cookies.subscribe([
-    {
-      name: 'session',  // Get change events for session-related cookies.
-      matchType: 'starts-with',  // Matches session_id, session-id, etc.
-    }
-  ]);
+  event.waitUntil(async () =>
```
inexorabletash: We need to add the `waitUntil` call in the spec example in https://wicg.github.io/cookie-store/#intro-monitor too.

explainer.md
```diff
@@ -169,11 +165,11 @@ The following code snippet uses the Cookie Store API instead, and does not jank
 the main thread.
 
 ```javascript
-document.getElementById('opt-out-button').addEventListener('click', () => {
+document.getElementById('opt-out-button').addEventListener('click', async () => {
   await cookieStore.set({ name: 'opt_out', value: '1',
                           expires: new Date('Wed, 1 Jan 2025 00:00:00 GMT') });
```
inexorabletash: FYI, this is implicitly doing a Date &rarr; Number conversion... which should do the right thing. But we may want to consider making this a number (e.g. `Date.now() + 1000*60*60*24*365`)  instead.

Also, 2025 will be in the past at some point.

explainer.md
pwnall: Done.

explainer.md
pwnall: Yup :)

This example is intended to match the `document.cookie` setter example above. I was fully aware of the implicit conversion, and wanted to (perhaps misleadingly?) point out that it should be easy to convert code that uses the `document.cookie` setter to the new API.

Do you think the concern above overrides the benefit of matching the `document.cookie` example? If so, do you think the `document.cookie` example should be updated to use a computed date as well, to match?

pwnall:
@inexorabletash Thank you for the quick and thorough review!

I'll merge this, and let's continue the discussion about the `expires:` value.

PR #134, opened 2020-04-07
Updating domain default definition

ayuishii:
### Description 
The following will cause different behaviors 
`domain : null` 
`domain: (new URL(self.location.href)).hostname`

Setting `null` which is the actual default means that it is locked toe the current domain and `host-only=flag` is set to false. 
Explicitly setting `domain` to `(new URL(self.location.href)).hostname` will set `host-only=true` and allow scope to span to subdomains as well.

### Reference 
[Cookie RFC](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.4)

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/134.html" title="Last updated on Apr 8, 2020, 8:35 PM UTC (5b3e740)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/134/7d1d875...5b3e740.html" title="Last updated on Apr 8, 2020, 8:35 PM UTC (5b3e740)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -263,8 +263,8 @@ await cookieStore.set({
   value: '1',
   expires: null,  // session cookie
 
-  // By default, cookies are scoped at the current domain.
-  domain: (new URL(self.location.href)).hostname,
+  // By default, domain is set to null which means the scope is locked at the current domain.
+  domain: null,
   path: '/',
 
   // Creates secure cookies by default on secure origins.
--- a/index.bs
+++ b/index.bs
@@ -355,8 +355,8 @@ await cookieStore.set({
   value: '1',
   expires: null,  // session cookie
 
-  // By default, cookies are scoped at the current domain.
-  domain: (new URL(self.location.href)).hostname,
+  // By default, domain is set to null which means the scope is locked at the current domain.
+  domain: null,
   path: '/',
 
   // Creates secure cookies by default on secure origins.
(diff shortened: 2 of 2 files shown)


ayuishii:
I tried updating the default domain section to help try remove some confusion. 
Open to suggestions on how to make this more clear. Thanks!

inexorabletash:
There's a corresponding example in the spec, and the algorithm doesn't account for this distinction:

https://wicg.github.io/cookie-store/#set-cookie-algorithm

> 3. If domain is not null, then run these steps:
>      1. If domain starts with U+002D (.), then return failure.
>      2. If host does not equal domain and host does not end with U+002D (.) followed by domain, then return failure.
> 4. Otherwise, if domain is null, then set domain to host.

Is the plan to update this to set `host-only`?

ayuishii:
Thanks for the comment. I made additional changes on the spec as you pointed out. 
I believe this behavior is already implemented [here](https://crrev.com/c/1102339) so just updating the spec to reflect its (expected) behavior.

inexorabletash:
The updated spec steps don't really work, because the [set a cookie](https://wicg.github.io/cookie-store/#set-cookie-algorithm) steps in the algorithm aren't operating on a cookie - they're converting arguments to pass into the RFC's [Storage Model "receives a cookie" algorithm](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.4) - that algorithm doesn't take an host-only-flag, but it does run similar logic to set it. I think we need to revise these steps to defer some of the logic to the RFC and e.g. not pass a `Domain` attribute rather than normalizing it before passing it in.

You could tackle that here, or just leave a TODO in the spec and we can follow up.

ayuishii:
Sorry for the delay, updated with commit to try and reflect the following suggested changes.

* Move the "Append Domain/domain (encoded) to attributes." into the "If domain is not null..." steps
* Drop "Otherwise, if domain is null, then set domain to host" entirely
* Move the "let attributes be a new list" step up before the first reference to attributes

Issue #121, opened 2020-01-23
Should `secure` default to true in CookieStoreSetOptions ?


inexorabletash:
Chrome's impl doesn't have `secure` default to true in `CookieStoreSetOptions`

The spec has `boolean secure = true`

What do we want here?

mkruisselbrink:
Related, per https://w3ctag.github.io/design-principles/#prefer-dict-to-bool we really shoulnd't have dictionary members that default to true.

pwnall:
Here's the reason for not having `secure` set to true in Blink, from [ToCanonicalCookie in cookie_store.cc](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/cookie_store/cookie_store.cc?q=ToCanonicalCookie)

Although the Cookie Store API spec always defaults the "secure" cookie attribute to true, we only default to true on cryptographically secure origins, where only secure cookies may be written, and to false otherwise, where only insecure cookies may be written.

As a result, `cookieStore.set("name", "value")` sets a cookie and `cookieStore.delete("name")` deletes a cookie on both `http://localhost` and secure origins, without having to specify `secure: false` on `http://localhost`.

pwnall:
Regarding the point raised by @mkruisselbrink -- there is a tradeoff between diverging from the TAG's guidelines and coming up with a different dictionary member, like `insecure: true`, to represent the [Secure attribute in RFC6265bis](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-5.3.5).

The current design chooses to diverge from the TAG guidelines. I think this results in slightly more readable code in most cases that have to deal with insecure cookies -- I'd guess that `secure: false` refers to the Secure attribute, but would have no idea what `insecure: true` means.

I'd be happy to get more feedback here. The implementation is fairly easy to change in this aspect.

inexorabletash:
Thanks for explaining! Specifying the blink behavior (which is effectively a tristate/optional Boolean) is possible, but it's going to be weird that it really only exists for localhost testing. 

We may want to highlight this for TAG feedback.

inexorabletash:
Also, we treat localhost as a secure context in most cases. I guess cookies are treated differently? Is that consistent across browsers? Is changing that possible? (Changing cookie semantics is trendy at the moment...)

mikewest:
[...]

PR #135, opened 2020-04-10
Remove `secure` flag from Modifications API 

ayuishii:
This change is based off https://github.com/WICG/cookie-store/issues/121, and removes `secure` option on Modifications API. Modifications API will only be able to write secure cookies.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/135.html" title="Last updated on May 15, 2020, 5:37 PM UTC (fe7817a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/135/7b808d7...fe7817a.html" title="Last updated on May 15, 2020, 5:37 PM UTC (fe7817a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -357,10 +357,7 @@ await cookieStore.set({
 
   // By default, domain is set to null which means the scope is locked at the current domain.
   domain: null,
-  path: '/',
-
-  // Creates secure cookies by default on secure origins.
-  secure: true,
+  path: '/'
 });
 ```
 
@@ -581,7 +578,6 @@ dictionary CookieStoreSetOptions {
   DOMTimeStamp? expires = null;
   USVString? domain = null;
   USVString path = "/";
-  boolean secure = true;
   CookieSameSite sameSite = "strict";
 };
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1224,11 +1216,11 @@ run the following steps:
         |host| does not end with U+002D (`.`) followed by |domain|,
         then return failure.
     1. [=list/Append=] \``Domain`\`/|domain| ([=encoded=]) to |attributes|.
-1. If |secure| is false and |name| starts with either "`__Secure-`" or "`__Host-`",
+1. If |name| starts with either "`__Secure-`" or "`__Host-`",
```
inexorabletash, line 1219: This seems like a behavior change - previously, this would fail with these prefixes only if _secure_ was false. Now it always fails with those prefixes. 

Was this previously wrong, or is this an intentional or accidental change?

index.bs
```diff
@@ -1272,10 +1264,9 @@ run the following steps:
     as long as it is in the past.
 
 1. Let |value| be the empty string.
-1. Let |secure| be true.
```
inexorabletash, line 1275: `|secure|` is still passed below as an argument (line 1288). It needs to be removed from the list of passed arguments.

index.bs
ayuishii: Ah sorry, you are right, this is an accidental change. 
But since |secure|  will be removed, should this line be removed?

ayuishii: Removed this statement since it will never be non-secure

PR #136, opened 2020-05-18
Add subtleties for modifying insecure cookies

ayuishii:
Adds a section about the subtleties for fetching and modifying insecure cookies with the API.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -575,6 +575,15 @@ const cookie = await cookieStore.get('session-id');
 console.log(cookie);
 ```
 
+### Modifying Insecure Cookies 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -575,6 +575,15 @@ const cookie = await cookieStore.get('session-id');
 console.log(cookie);
 ```
 
+### Modifying Insecure Cookies 
+The API will be able to fetch insecure cookies, however will only be able to modify secure cookies. This will mean that when modifying an insecure cookie with the API, the insecure cookie will automatically be modified to secure.
```
inexorabletash, line 579: Phrasing suggestions:

"...however will..." &rarr; "...but will..."

"...modified to..." &rarr; "...changed to..."

ayuishii: Sounds good!

inexorabletash:
Thanks!

Issue #139, opened 2020-05-28
Update `set()` method to use CookieInit


ayuishii:
Per recommendation from kenchris@ via [TAG discussion](https://github.com/w3ctag/design-reviews/issues/469#issuecomment-633879420)

Change `set()` method to be as follows 

**Before**
```
 Promise<void> set(USVString name, USVString value,
                    optional CookieStoreSetOptions options = {});
 Promise<void> set(CookieStoreSetExtraOptions options);

dictionary CookieStoreSetOptions {
  DOMTimeStamp? expires = null;
  USVString? domain = null;
  USVString path = "/";
  CookieSameSite sameSite = "strict";
};

dictionary CookieStoreSetExtraOptions : CookieStoreSetOptions {
  required USVString name;
  required USVString value;
};
```

**After**
```
  Promise<void> set(USVString name, USVString value);
  Promise<void> set(CookieListItemInit cookieListItemInit);

dictionary CookieListItemInit {
  required USVString name;
  USVString value;
  USVString? domain = null;
  USVString path = "/";
  DOMTimeStamp? expires = null;
  CookieSameSite sameSite = "strict";
};
```

inexorabletash:
Maybe rename `CookieListItemInit` to just `CookieInit` ?

ayuishii:
`CookieInit` SGTM 👍

ayuishii:
Spec change and implementation complete, thanks!

PR #141, opened 2020-06-02
Update `set()` method to use CookieInit

ayuishii:
Github Issue: https://github.com/WICG/cookie-store/issues/139

Change `set()` method to be as follows 

**Before**
```
Promise<void> set(USVString name, USVString value,
                    optional CookieStoreSetOptions options = {});
Promise<void> set(CookieStoreSetExtraOptions options);

dictionary CookieStoreSetOptions {
  DOMTimeStamp? expires = null;
  USVString? domain = null;
  USVString path = "/";
};

dictionary CookieStoreSetExtraOptions : CookieStoreSetOptions {
  required USVString name;
  required USVString value;
};
```

**After**
```
Promise<void> set(USVString name, USVString value);
Promise<void> set(CookieInit cookieInit);

dictionary CookieInit {
  required USVString name;
  USVString value;
  USVString? domain = null;
  USVString path = "/";
  DOMTimeStamp? expires = null;
};
```
Usage examples:
```js
cookieStore.set('cookie-name', 'cookie-value');
cookieStore.set({ name: 'cookie-name', value: 'cookie-value', path: '/' });
```

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/141.html" title="Last updated on Jun 3, 2020, 5:51 PM UTC (7dff5dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/141/5afd965...7dff5dc.html" title="Last updated on Jun 3, 2020, 5:51 PM UTC (7dff5dc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -334,7 +334,7 @@ if (cookie.secure)
 Both [=documents=] and [=service workers=] access the same modification API, via the
 {{Window/cookieStore}} property on the [[#globals|global object]].
 
-Cookies are created or modified (written) using the {{CookieStore/set(options)|set()}} method.
+Cookies are created or modified (written) using the {{CookieStore/set(name, value)|set()}} method.
 
 <div class=example>
 Write a cookie:
@@ -347,7 +347,7 @@ try {
 }
 ```
 
-The {{CookieStore/set(options)|set()}} call above is a shorthand for the following:
+The {{CookieStore/set(name, value)|set()}} call above is shorthand for using an options dictionary, as follows:
 
 ```js
 await cookieStore.set({
@@ -577,7 +577,7 @@ enum CookieSameSite {
 
 dictionary CookieInit {
   required USVString name;
-  USVString value;
+  required USVString value;
   DOMTimeStamp? expires = null;
   USVString? domain = null;
   USVString path = "/";
--- a/index.bs
+++ b/index.bs
@@ -125,11 +125,12 @@ What if you could instead write:
 
 <div class=example>
 ```js
+const one_day_ms = 24 * 60 * 60 * 1000;
 cookieStore.set(
   {
     name: '__Secure-COOKIENAME',
     value: 'cookie-value',
-    expires: Date.now() + 24*60*60*1000,
+    expires: Date.now() + one_day_ms,
     domain: 'example.org'
   }).then(function() {
     console.log('It worked!');
--- a/index.bs
+++ b/index.bs
@@ -591,15 +591,9 @@ dictionary CookieStoreDeleteOptions {
   USVString path = "/";
 };
 
-dictionary CookieListItem {
-  required USVString name;
-  USVString value;
-  USVString? domain = null;
-  USVString path = "/";
-  DOMTimeStamp? expires = null;
+dictionary CookieListItem : CookieInit {
   boolean secure = true;
-  CookieSameSite sameSite = "strict";
-};
+}
 
 typedef sequence<CookieListItem> CookieList;
 </xmp>
--- a/index.bs
+++ b/index.bs
@@ -126,9 +126,9 @@ What if you could instead write:
 <div class=example>
 ```js
 cookieStore.set(
-  '__Secure-COOKIENAME',
-  'cookie-value',
   {
+    name: '__Secure-COOKIENAME',
+    value: 'cookie-value',
     expires: Date.now() + 24*60*60*1000,
     domain: 'example.org'
   }).then(function() {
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -347,7 +347,7 @@ try {
 }
 ```
 
-The {{CookieStore/set(name, value, options)|set()}} call above is a shorthand for the following:
+The {{CookieStore/set(options)|set()}} call above is a shorthand for the following:
```
inexorabletash, line 350: I think the link here should be to `{{CookieStore/set(name, value)|set()}}` ?

index.bs
```diff
@@ -334,7 +334,7 @@ if (cookie.secure)
 Both [=documents=] and [=service workers=] access the same modification API, via the
 {{Window/cookieStore}} property on the [[#globals|global object]].
 
-Cookies are created or modified (written) using the {{CookieStore/set(name, value, options)|set()}} method.
+Cookies are created or modified (written) using the {{CookieStore/set(options)|set()}} method.
```
inexorabletash, line 337: Given that the following example uses name, value shouldn't this be {{CookieStore/set(options)|set()}} ?

index.bs
inexorabletash: And maybe it should say "shorthand for using an options dictionary, as follows:" ?

index.bs
```diff
@@ -385,8 +385,10 @@ Deleting a cookie by changing the expiry date:
 ```js
 try {
   const one_day_ms = 24 * 60 * 60 * 1000;
```
inexorabletash, line 387: Not new: maybe we should use const declaration in all of the examples where we're currently using this calculation inline.

index.bs
```diff
@@ -574,18 +575,15 @@ enum CookieSameSite {
   "none"
 };
 
-dictionary CookieStoreSetOptions {
+dictionary CookieInit {
```
inexorabletash, line 578: Hmmm.. wasn't the intent of the change to merge this with the CookieListItem dictionary, or was I mistaken?

index.bs
```diff
@@ -772,12 +766,12 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
...
     1. Let |r| be the result of running [=set a cookie=] with
         |url|,
-        |options|' {{CookieStoreSetExtraOptions/name}} dictionary member,
-        |options|' {{CookieStoreSetExtraOptions/value}} dictionary member,
-        |options|' {{CookieStoreSetOptions/expires}} dictionary member,
-        |options|' {{CookieStoreSetOptions/domain}} dictionary member,
-        |options|' {{CookieStoreSetOptions/path}} dictionary member, and
-        |options|' {{CookieStoreSetOptions/sameSite}} dictionary member.
+        |options|' {{CookieInit/name}} dictionary member,
+        |options|' {{CookieInit/value}} dictionary member,
```
inexorabletash, line 770: `value` is not a required dictionary member now, but it's not an optional argument for the *set a cookie* algorithm, so we need to define what's passed here.

Can `value` be made required, or will that prevent something?

index.bs
ayuishii: Oh I see, I think I misunderstood. The only difference between CookieListItem and CookieInit is that CookieInit does not have `secure`. So should this be something like this?
```
dictionary CookieInit {
  required USVString name;
  USVString value;
  USVString? domain = null;
  USVString path = "/";
  DOMTimeStamp? expires = null;
};

dictionary CookieListItem : CookieInit {
  boolean secure = true;
}

index.bs
ayuishii: Oh, good point.. I think this can be required. I will update.

index.bs
ayuishii: Ah you are right, thanks for the catch.

index.bs
ayuishii: SG, thanks for the catch.

index.bs
ayuishii: I think this is the only place this calculation is being used

inexorabletash: ` expires: Date.now() - 24 * 60 * 60 * 1000 });`  up above?

ayuishii: Ooh, I missed that. Thanks for the clarification!

index.bs
inexorabletash: That LGTM. Maybe the `secure` member wasn't apparent to kenneth. 

So once this all lands we may want to ping and say "here's what we ended up with" (but we can land and then ask)

ayuishii: SG 👍 , thanks!

PR #143, opened 2020-06-03
Remove `required` from CookieInit value

ayuishii:
I bumped into an issue while trying to implement `required` for value, and thought it could be discussed here. 

Currently we spec that an `undefined` value is set for CookieListItem for a deleted change event CookieListItem described [here](https://wicg.github.io/cookie-store/#process-changes). Tested and verified [here](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/external/wpt/cookie-store/cookieStore_event_delete.tentative.https.window.js;l=19?originalUrl=https:%2F%2Fcs.chromium.org%2F).

Would be acceptable to have this return an empty string instead of `undefined` or maybe to have a separate dict to add `required` to value string just to set(). 

Let me know if a different forum would be better.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/143.html" title="Last updated on Jun 5, 2020, 8:24 PM UTC (0b3f3b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/143/d6308dd...0b3f3b8.html" title="Last updated on Jun 5, 2020, 8:24 PM UTC (0b3f3b8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -592,7 +592,7 @@ dictionary CookieStoreDeleteOptions {
 };
 
 dictionary CookieListItem {
-  required USVString name;
+  USVString name;
   USVString value;
   USVString? domain;
   USVString path;
--- a/index.bs
+++ b/index.bs
@@ -578,7 +578,7 @@ enum CookieSameSite {
 
 dictionary CookieInit {
   required USVString name;
-  required USVString value;
+  USVString value;
   DOMTimeStamp? expires = null;
   USVString? domain = null;
   USVString path = "/";
--- a/index.bs
+++ b/index.bs
@@ -578,7 +578,7 @@ enum CookieSameSite {
 
 dictionary CookieInit {
   required USVString name;
-  USVString value;
+  required USVString value;
   DOMTimeStamp? expires = null;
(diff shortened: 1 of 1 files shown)


ayuishii:
Thanks for the input @domenic !
I agree with the clarity and the benefit of being able to remove defaults.
I will update the PR, but let me know what you think @inexorabletash .

index.bs
```diff
@@ -591,8 +591,14 @@ dictionary CookieStoreDeleteOptions {
   USVString path = "/";
 };
 
-dictionary CookieListItem : CookieInit {
-  boolean secure = true;
+dictionary CookieListItem {
+  required USVString name;
```
inexorabletash, line 595: Since this is an output-only dictionary, `required` doesn't have any meaning. Remove it?

ayuishii: Removed

inexorabletash:
LGTM!

ayuishii:
Thanks for the reviews!

PR #144, opened 2020-06-08
Revert "Remove optional from CookieStoreGetOptions"

ayuishii:
IDL rules disallows required dictionaries that don't have any required fields.
https://github.com/w3c/webidl2.js/blob/gh-pages/README.md#errors

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -219,8 +219,9 @@ In Document contexts, `await cookieStore.getAll()` is an equivalent of
 `document.cookie`.
 
 In other words, `get` and `getAll` take the same arguments, which can be
-* a name
-* a dictionary of options (optional for `getAll`)
+* an optional dictionary of options
+* a name and an optional dictionary of options; in this case, the bag must not
+  contain the `name` properties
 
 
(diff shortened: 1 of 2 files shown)


domenic:
FYI, similar situations have come up a few times in the past, where the spec want the semantics of "one of the dictionary members is required, but not any particular one". A notable one is https://dom.spec.whatwg.org/#dom-mutationobserver-observe . The solution used there was to do a validity check in prose, that ensured the dictionary wasn't empty, even though IDL declared it optional.

It's not 100% clear to me what the spec does in this case, but I wanted to offer that as an option.

ayuishii:
Thanks for the providing the FYI @domenic !

I don't have enough experience to have developed much of an instinct with these things yet, but my personal thought is that maybe keeping _optional_ is ok for this case. Since Kenneth's main concern on the TAG review seemed to be about a TypeError being thrown when being passed an empty option while allowing for it in the IDL, which is not the case, but will instead return the first cookie or `null` if empty. I do like the idea of keeping it flexible without requiring any of the fields although unsure of the value it would provide to do so...

PR #145, opened 2020-06-10
Add missing semicolon after CookieListItem dictionary

foolip:
Discovered in https://github.com/web-platform-tests/wpt/pull/23957.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -599,7 +599,7 @@ dictionary CookieListItem {
   DOMTimeStamp? expires;
   boolean secure;
   CookieSameSite sameSite;
-}
+};
 
 typedef sequence<CookieListItem> CookieList;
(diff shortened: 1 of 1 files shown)


foolip:
This was introduced in https://github.com/WICG/cookie-store/pull/141, and Bikeshed does in fact complain very loudly about this if run locally:
> FATAL ERROR: IDL SYNTAX ERROR LINE: 56 - expected ";"

PR Preview always tries to generate some output, which makes sense, but it does mean that errors are missed.

foolip:
I've filed https://github.com/tobie/pr-preview/issues/68 about what could have avoided this.

tobie:
Opened https://github.com/WICG/cookie-store/pull/146 to fix this.

Issue #138, opened 2020-05-28
Document "clear cookies" behavior


ayuishii:
"Clear cookies" for an origin _needs_ to wipe all storage for that origin; including service workers. 

Let's make a non-normative, but explicit note on this behavior for "clear cookies" in the spec.

per [TAG discussion](https://github.com/w3ctag/design-reviews/issues/469#issuecomment-633875799)

PR #148, opened 2020-06-22
Document Clear Cookies

ayuishii:
For issue: https://github.com/WICG/cookie-store/issues/138

Adding a section to describe the expected browser behavior for when the user initiates "clear cookies" for an origin.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1435,7 +1435,7 @@ Likewise, today reading {{Document/cookie|document.cookie}} in contexts where sc
 ## Clear cookies ## {#clear-cookies}
 <!-- ============================================================ -->
 
-When a user clears cookies for an origin, the browser needs to wipe all storage for that origin; including service workers. This is to prevent service workers from restoring any user identifiers in persisted storage that outlives "clear cookies".
+When a user clears cookies for an origin, the browser needs to wipe all storage for that origin; including service workers. This is to prevent websites from restoring any user identifiers in persistent storage after a user initiates "clear cookies".
 
 <!-- ====================================================================== -->
 # Acknowledgements # {#acknowledgements}
--- a/index.bs
+++ b/index.bs
@@ -1441,7 +1441,7 @@ Likewise, today reading {{Document/cookie|document.cookie}} in contexts where sc
 
 *This section is non-normative.*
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1431,6 +1431,12 @@ Today writing to {{Document/cookie|document.cookie}} in contexts where script-in
 
 Likewise, today reading {{Document/cookie|document.cookie}} in contexts where script-initiated cookie-reading is disallowed typically returns an empty string. However, a cooperating web server can verify that server-initiated cookie-writing and cookie-reading work and report this to the script (which still sees empty string) and the script can use this information to infer that script-initiated cookie-reading is disallowed.
 
+<!-- ============================================================ -->
+## Clear cookies ## {#clear-cookies}
```
inexorabletash, line 1435: Would it make sense to include this in a (new) "Privacy considerations" section?

("Security" should get renamed "Security considerations", and we should eventually add Accessibility and Internationalization considerations sections as well)

index.bs
```diff
@@ -1431,6 +1431,12 @@ Today writing to {{Document/cookie|document.cookie}} in contexts where script-in
...
+<!-- ============================================================ -->
+
```
inexorabletash, line 1437: This should be called out as "non-normative" (e.g. just add `*This section is non-normative.*` here)

index.bs
```diff
@@ -1431,6 +1431,12 @@ Today writing to {{Document/cookie|document.cookie}} in contexts where script-in
...
+When a user clears cookies for an origin, the browser needs to wipe all storage for that origin; including service workers. This is to prevent websites from restoring any user identifiers in persistent storage after a user initiates "clear cookies".
```
inexorabletash, line 1438: replace "browser" with "user agent"

maybe reword "a user initiates "clear cookies"" with "a user initiates the action" ?

You may want to look at https://w3c.github.io/webappsec-clear-site-data/#clear-dom for some more precise terminology around "all storage for that origin", although that spec itself still needs to integrate with Storage about what it really means. Specifically, you could write "also clear DOM-accessible storage for that origin"

index.bs
ayuishii: Thanks for pointing this out!

index.bs
ayuishii: Thanks for the quick review! Sounds good, I added the text about DOM-accessible storage and added a link with the regular mark down linking format. I found one example of this but wasn't sure if its the most appropriate way to link across specs. Let me know if there is a better way.

inexorabletash: Hmmm... I might leave out the link for now, and/or maybe add an inline `Issue: ....` block talking about how this should link to a canonical definition, maybe in Storage.

ayuishii:
Removed the link, thanks for the review @inexorabletash !

Issue #128, opened 2020-03-20
Is `matchType` necessary?


mikewest:
The explainer presents the Google+ API as a use case for the `matchType: 'starts-with'` option. It's not clear to me that that's something we can reasonably point to today. If we dropped it from `CookieStoreGetOptions`, we could simplify the `Window` exposure by dropping the `CookieStoreGetOptions` variants of `get()` and `getAll()` (as `url` is only meaningful in Service Workers).

ayuishii:
Thanks for the feedback @mikewest !
I think this would be a good candidate for getting feedback and observing how this is used during Origin Trial. I can put in place some metrics to do so. Does this sound ok?

domenic:
Where did the origin trial feedback end up in this regard?

ayuishii:
Looking through our metrics and connecting with our partner, looks like we don't have a strong case to keep `matchType`. 
I've created a [bug](https://github.com/WICG/cookie-store/issues/128) to track removal and will update the spec / explainer on this change.

PR #167, opened 2020-09-03
Remove `matchType`

ayuishii:
Discussion: https://github.com/WICG/cookie-store/issues/128
Implementation tracking: https://crbug.com/1124497

***
<a href="https://pr-preview.s3.amazonaws.com/ayuishii/cookie-store/pull/167.html" title="Last updated on Sep 3, 2020, 9:42 PM UTC (28d876a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/167/035ad0f...ayuishii:28d876a.html" title="Last updated on Sep 3, 2020, 9:42 PM UTC (28d876a)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -204,8 +204,7 @@ try {
 ```javascript
 try {
   const cookies = await cookieStore.getAll({
-    name: 'session_',
-    matchType: 'starts-with',
+    name: 'session_id'
   });
   for (const cookie of cookies)
     console.log(`Result: ${cookie.name} = ${cookie.value}`);
@@ -386,8 +385,7 @@ self.addEventListener('activate', (event) => {
 
     await self.registration.cookies.subscribe([
       {
-        name: 'session',  // Get change events for session-related cookies.
-        matchType: 'starts-with',  // Matches session_id, session-id, etc.
+        name: 'session_id',  // Get change events for cookies named session_id.
       }
     ]);
   });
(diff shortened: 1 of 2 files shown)


inexorabletash (approved):
lgtm...

We may want to add a note somewhere that {name} is passed as an options dict rather than just as a string argument to allow future flexibility. It might be good to get domenic's feedback on that, though.

ayuishii:
@inexorabletash Just to understand better, do you mean removing `get(USVString name)` variant and only accepting options for future flexibility and keeping the CookieStoreGetOptions variant available for `Window`? 

(Sorry, I noticed this PR itself is incomplete as I forgot to add simplification of `Window` exposure suggested in the [issue](https://github.com/WICG/cookie-store/issues/128_.))

/cc @domenic

domenic:
I think the simple `get(USVString name)` variant is a good thing to keep. At the same time, you need to keep the `get(options)` variant for service workers. #128 suggest un-exposing the options variant in `Window`, but I'm not sure about that... letting people specify an explicit match URL might be useful even in Windows?

The examples should probably be updated to use the simpler string version, though.

domenic:
> I'm not sure about that... letting people specify an explicit match URL might be useful even in Windows?

I've thought about this more and I'd like to strengthen my "I'm not sure" :). I'm really not sure whether it's useful; maybe it's safer to just leave it out. I trust you all as domain experts more than my judgment on this one.

ayuishii:
Thanks for the feedback!

Personally I think having both options, `get(USVString name)` & `get(options)` would be nice in `Window`. Both for simplicity for the name option, and also flexibility of the options version..but curious to hear your thoughts @inexorabletash ?

inexorabletash:
Agreed re: flexibility. And I don't think anyone will really complain. Per [priority of constituencies](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies):

Users (developers): unlikely to notice redundancy between `get(name)` & `get({name})`
Implementors: simpler to support both in both window & sw contexts
Spec authors: simpler to have single CookieStore definition

Glad we've had the discussion though!

ayuishii:
Awesome, thanks for the discussion everyone, this was super helpful!
I updated `getAll()` example to use simple string version.

Issue #159, opened 2020-09-01
Task queuing needs a task source (and document)


domenic:
> Queue a task to fire a change event named "change" with changes at window’s CookieStore on window’s responsible event loop.

This should use [queue a global task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-global-task) given _window_, to pass along the document when appropriate. But that's pretty minor (and easy for the reader to infer).

The trickier question is what task source to use for the queued task. I'm curious which one Chromium uses? If it's kMiscPlatformAPI, then that's bad and we need to pick a new one.

inexorabletash:
It's `kMiscPlatformAPI` - how about `kDOMManipulation` like BroadcastChannel and Local/SessionStorage ?

ayuishii:
Updated spec to use `kDOMManipulation`. Will track implementation [here](https://crbug.com/1128057).

PR #171, opened 2020-09-08
Fix task queuing

ayuishii:
Discussion: https://github.com/WICG/cookie-store/issues/159

***
<a href="https://pr-preview.s3.amazonaws.com/ayuishii/cookie-store/pull/171.html" title="Last updated on Sep 9, 2020, 11:20 PM UTC (0f772c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/171/1d2a91f...ayuishii:0f772c6.html" title="Last updated on Sep 9, 2020, 11:20 PM UTC (0f772c6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1233,9 +1233,8 @@ To <dfn>process cookie changes</dfn>, run the following steps:
     1. Let |url| be |window|'s [=creation URL=].
     1. Let |changes| be the [=observable changes=] for |url|.
     1. If |changes| [=set/is empty=], then [=continue=].
-    1. [=Queue a global task=] to
-        [=fire a change event=] named "`change`" with |changes| at |window|'s {{CookieStore}}
-        on |window|'s [=responsible event loop=].
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |window| to
+        [=fire a change event=] named "`change`" with |changes| at |window|'s {{CookieStore}}.
 
 1. For every [=service worker registration=] |registration|, run the following steps:
     1. Let |changes| be a new [=/set=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1233,7 +1233,7 @@ To <dfn>process cookie changes</dfn>, run the following steps:
     1. Let |url| be |window|'s [=creation URL=].
     1. Let |changes| be the [=observable changes=] for |url|.
     1. If |changes| [=set/is empty=], then [=continue=].
-    1. [=Queue a task=] to
+    1. [=Queue a global task=] to
```
inexorabletash, line 1236: We probably need @domenic to review, but...

When specifying "global task" we don't need to pull out the event loop from window (that's what "global task" does, per https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-global-task). We also want to follow the phrasing:

Queue a global task on XYZ task source given global to steps.

so e.g.

Queue a global task on XYZ task source given window to fire a change event named "change" with changes at window's CookieStore.

And then.. which task source? Storage and BroadcastChannel use the DOM manipulation task source so I assume it would be the same...? As @domenic asks on #159 we probably need to update the code. I'd go with DOM manipulation task source.

inexorabletash (changes requested):

index.bs
ayuishii: Thanks for explaining @inexorabletash ! I've updated the wording and changed to use DOM manipulation task. 
Will ask domenic for additional review when he's back.

inexorabletash (approved):

domenic (approved):
LGTM!

The other natural choice would be the network task source. It seems like it kind of depends on whether you consider this a "network API" or a "storage API".

It's hard to think of a case where this would really matter. I guess it's possible that when choosing DOM manipulation task source, the `"change"` event could fire before the corresponding `fetch().then(onResponse)` handler gets called, if the browser decides to prioritize DOM manipulation (cookie change event) ahead of networking (`onResponse` handler calling). But that seems OK, and I imagine you could cook up similar scenarios with BroadcastChannel.

ayuishii:
Sounds good, thanks for the reviews & explanation! I will try with DOM manipulation task, and will track implementation [here](https://crbug.com/1128057).

Issue #93, opened 2018-10-23
Remove set(CookieStoreSetExtraOptions options)?


jakearchibald:
This creates a situation where we have an options object with required options.

Sure, it means the output of `get` can be passed to `set` (after modifications), but this is racey, and not something we should encourage.

You can reduce some of the racing with web locks, but since responses can also set cookies, it isn't a total solution.

jakearchibald:
`delete(CookieStoreDeleteOptions options)` has the same problem. Options shouldn't be required.

domenic:
It seems like the current draft still keeps this options API. Should this be closed? If so, it'd be good to at least explain the potential race and danger.

ayuishii:
Sounds good, will look into expanding this [section](https://github.com/WICG/cookie-store/blob/main/explainer.md#subtleties) in the explainer to talk about the dangers.

PR #176, opened 2020-09-11
Add explanation about read/write data race

ayuishii:
Discussion: https://github.com/WICG/cookie-store/issues/93

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -572,6 +572,24 @@ const cookie = await cookieStore.get('session-id');
 console.log(cookie);
 ```
 
+#### Read / Write Race
+
+Cookie store operations are not an atomic read-modify-write transaction. Failure to await during an operation can cause unexpected results as the following.
+
+```javascript
+function increment_cookie(name) {
+  const cookie = await cookieStore.get(name);
+  cookie.value = parseInt(cookie.value) + 1;
+  await cookieStore.set(cookie);
+  return cookie;
+}
+
+await cookieStore.set('cookie-name', 1);
+const cookie1 = increment_cookie('cookie-name');  // Missing "await".
+const cookie2 = increment_cookie('cookie-name');  // Missing "await".
+// cookie1.value and cookie2.value may have the same value 2. 
+```
+
 ### Modifying Insecure Cookies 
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -572,6 +572,24 @@ const cookie = await cookieStore.get('session-id');
 console.log(cookie);
 ```
 
+#### Read / Write Race
+
+Cookie store operations are not an atomic read-modify-write transaction. Failure to await during an operation can cause unexpected results as the following.
+
+```javascript
+function increment_cookie(name) {
```
pwnall, line 580: `async function` because you use await

`incrementCookie` because JavaScript

explainer.md
```diff
@@ -572,6 +572,24 @@ const cookie = await cookieStore.get('session-id');
...
+  const cookie = await cookieStore.get(name);
+  cookie.value = parseInt(cookie.value) + 1;
+  await cookieStore.set(cookie);
+  return cookie;
+}
+
+await cookieStore.set('cookie-name', 1);
+const cookie1 = increment_cookie('cookie-name');  // Missing "await".
```
pwnall, line 588: The values here are cookie1_promise and cookie2_promise

How about

```js
await Promise.all([
  incrementCookie('cookie-name'),
  incrementCookie('cookie-name'),
]);
const cookie = await cookieStore.get('cookie-name');
console.log(cookie.value);  // will likely be '2' not '3'
```

explainer.md
```diff
@@ -572,6 +572,24 @@ const cookie = await cookieStore.get('session-id');
 console.log(cookie);
 ```
 
+#### Read / Write Race
+
+Cookie store operations are not an atomic read-modify-write transaction. Failure to await during an operation can cause unexpected results as the following.
```
pwnall, line 577: The ability of passing the return value of `cookieStore.set()` to `cookieStore.get()` may lead to code that reads like an atomic read-modify-write operation, when it reality it's not.

pwnall (changes requested):
Thank you very much for taking this on!

Here are a couple of suggestions for changes.

I found that testing my examples in DevTools helps remove small JS bugs, leading to more credible explainers and specs.

ayuishii:
Thanks for the review! And pointing out all the issues.
Will be better about verifying in the future...

pwnall (approved):
LGTM, thank you for the quick revisions!

Issue #180, opened 2020-09-16
Algorithms for CookieStoreManager treat |this| as a service worker registration


mkruisselbrink:
The algorithms for the methods on CookieStoreManager in https://wicg.github.io/cookie-store/#CookieStoreManager all use `Let registration be this.`, and then operate on `registration`as if it is a `service worker registration`. But `this` in this context would be the actual `CookieStoreManager` object.

The `CookieStoreManager` object probably needs to have an associated service worker registration, which the algorithms can then use instead.

ayuishii:
Thanks for the guidance!
Issue addressed here: https://github.com/WICG/cookie-store/pull/181

PR #181, opened 2020-09-16
Associated Service Worker Registration for CookieStoreManager

ayuishii:
Issue: https://github.com/WICG/cookie-store/issues/180

***
<a href="https://pr-preview.s3.amazonaws.com/ayuishii/cookie-store/pull/181.html" title="Last updated on Sep 17, 2020, 12:36 AM UTC (b1ca92d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/181/450c023...ayuishii:b1ca92d.html" title="Last updated on Sep 17, 2020, 12:36 AM UTC (b1ca92d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -915,15 +915,19 @@ The <dfn method for=CookieStoreManager>unsubscribe(|subscriptions|)</dfn> method
 ## The {{ServiceWorkerRegistration}} interface ## {#ServiceWorkerRegistration}
 <!-- ============================================================ -->
 
-The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}} via {{cookies}} which provides the interface for subscribing to cookie changes.
+The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}} via {{ServiceWorkerRegistration/cookies}} which provides the interface for subscribing to cookie changes.
 
 <xmp class=idl>
 [Exposed=(ServiceWorker,Window)]
 partial interface ServiceWorkerRegistration {
-  readonly attribute CookieStoreManager cookies;
+  [SameObject] readonly attribute CookieStoreManager cookies;
 };
 </xmp>
 
+Each {{ServiceWorkerRegistration}} has an associated {{CookieStoreManager}} object.
+The {{CookieStoreManager}}'s [=CookieStoreManager/registration=] is equal to the {{ServiceWorkerRegistration}}'s [=service worker registration=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -913,7 +915,7 @@ The <dfn method for=CookieStoreManager>unsubscribe(|subscriptions|)</dfn> method
 ## The {{ServiceWorkerRegistration}} interface ## {#ServiceWorkerRegistration}
 <!-- ============================================================ -->
 
-The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}}.
+The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}} via {{cookies}} which provides the interface for subscribing to cookie changes.
 
 <xmp class=idl>
 [Exposed=(ServiceWorker,Window)]
```
mkruisselbrink, line 921: kind of separate to this PR, but cookies should probably be [SameObject]. And the cookies attribute should normatively declare what it returns somewhere, i.e. something like this:

```
Each {{ServiceWorkerRegistration}} has an associated {{CookieStore}} object.
The {{CookieStore}}'s [=CookieStore/registration=] is equal to the {{ServiceWorkerRegistration}}'s [=ServiceWorkerRegistration/service worker registration=].

The {{ServiceWorkerRegistration/cookies}} [=getter steps=] are to return [=this=]'s associated {{CookieStore}} object.
```

ayuishii: Thanks for the guidance! I made the change to SameObject and added the normative bit below. Made some changes to get the links to work, but let me know if I've missed anything.

mkruisselbrink: Huh, I see. Technically it really should be [=ServiceWorkerRegistration/service worker registration=], but I guess that term isn't exported by the service worker spec, so what you have now seems good enough.
(as background, [=service worker registration=] would link to https://w3c.github.io/ServiceWorker/#dfn-service-worker-registration, while what we really want here is https://w3c.github.io/ServiceWorker/#serviceworkerregistration-service-worker-registration, but as said that term isn't exported. It's not like it isn't clear what we're talking about so the current text seems fine to me).

mkruisselbrink (approved):

index.bs
ayuishii: I see, just out of curiosity how can you tell if a term has been exported or not?

ayuishii:
Thank you for the review!

Issue #188, opened 2020-12-01
Require leading / in paths for modifications


inexorabletash:
Explainer says that modifications (set, delete) require a leading `/` in paths:
* https://github.com/WICG/cookie-store/blob/main/explainer.md#paths

Tests validates this:
* https://github.com/web-platform-tests/wpt/tree/master/cookie-store/cookieStore_delete_arguments.tentative.https.any.js
* https://github.com/web-platform-tests/wpt/blob/master/cookie-store/cookieStore_set_arguments.tentative.https.any.js

But the spec seemingly doesn't have anything requiring this:
* https://wicg.github.io/cookie-store/#set-cookie-algorithm

Add it to the spec?

inexorabletash:
This was fixed.

PR #192, opened 2020-12-15
Update `Path` Algorithm to match Explainer & Tests

ayuishii:
A leading `/` is required in paths for setting & deleting a cookie.
This is part of the explainer and has tests to validate this, but was not part of the spec.
This change adds this to the spec.

https://github.com/WICG/cookie-store/issues/188

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/192.html" title="Last updated on Apr 26, 2021, 8:26 PM UTC (fceaff6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/192/62089ad...fceaff6.html" title="Last updated on Apr 26, 2021, 8:26 PM UTC (fceaff6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1110,8 +1110,8 @@ run the following steps:
     1. [=list/Append=] \``Domain`\`/|domain| ([=encoded=]) to |attributes|.
 1. If |expires| is given, then [=list/append=] \``Expires`\`/|expires| ([=date serialized=]) to |attributes|.
 1. If |path| is not null, then run these steps:
-    1. If |path| does not start with U+0002F (`/`), then return failure.
-    1. If |path| does not end with U+0002F (`/`), append U+0002F (`/`) to |path|.
+    1. If |path| does not start with U+002F (`/`), then return failure.
+    1. If |path| does not end with U+002F (`/`), then append U+002F (`/`) to |path|.
 1. [=list/Append=] \``Path`\`/|path| ([=encoded=]) to |attributes|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1109,6 +1109,9 @@ run the following steps:
         then return failure.
     1. [=list/Append=] \``Domain`\`/|domain| ([=encoded=]) to |attributes|.
 1. If |expires| is given, then [=list/append=] \``Expires`\`/|expires| ([=date serialized=]) to |attributes|.
+1. If |path| is not null, then run these steps:
+    1. If |path| does not start with U+0002F (`/`), then return failure.
```
inexorabletash, line 1113: U+nnnn is usually written with zero padding to bring it to four hex digits minimum, so this should be U+002F

index.bs
```diff
@@ -1109,6 +1109,9 @@ run the following steps:
...
+    1. If |path| does not end with U+0002F (`/`), append U+0002F (`/`) to |path|.
```
inexorabletash, line 1114: style: I've been trying to use "if" paired with "then" explicitly. Add "then" ?

(The spec isn't 100% consistent, though.)

inexorabletash (changes requested):
Two minor nits, which apply to multiple parts of the PR

(sorry I split them into multiple comments!)

inexorabletash (approved):

PR #195, opened 2021-05-01
Fix example with setting the cookie

MeFoDy:
@lahmatiy noticed in reply to my [small blog post](https://mefody.dev/chunks/cookie-store-api/) that there are two places the code snippet can become better :)


MeFoDy:
Duplicates #196, I close it.

PR #196, opened 2021-05-01
Fix and simplify cookie writing example

lahmatiy:
1. `URL` interface has no `domain` property but `host`
2. `new URL(self.location.href)` is not needed since `self.location` has similar interface as `URL` (see https://html.spec.whatwg.org/multipage/history.html#location, https://html.spec.whatwg.org/multipage/workers.html#workerlocation)


inexorabletash:
Thanks!

PR #200, opened 2021-08-05
Add char set / size checks to 'Set a Cookie' section

### LINKED ISSUE #199: Add character set and length restrictions from RFC6265bis
We should update the 'Set a cookie' section to mention the character set limitations and length restrictions present in the latest version of RFC6265bis. PR incoming!

recvfrom:
Fixes #199 

This PR attempts to add in the size checks and the control character checks from the latest version of RFC6265bis.

Since the RFC specifies the size limits in octets, proper encoding of the applicable parameters must occur so that the size checks are computed correctly given JavaScript strings.

Please let me know if you have any feedback or suggestions for improving this PR. Thank you!

***
<a href="https://pr-preview.s3.amazonaws.com/recvfrom/cookie-store/pull/200.html" title="Last updated on Aug 6, 2021, 9:37 PM UTC (853dcfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/200/77f41a8...recvfrom:853dcfa.html" title="Last updated on Aug 6, 2021, 9:37 PM UTC (853dcfa)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -429,6 +429,10 @@ Per [[RFC6265bis#section-5.4|Cookies: HTTP State Management Mechanism §Storage 
 
 A cookie is <dfn>script-visible</dfn> when it is in-scope and does not have the `HttpOnly` cookie flag. This is more formally enforced in the processing model, which consults [[RFC6265bis#section-5.5|Cookies: HTTP State Management Mechanism §The Cookie Header]] at appropriate points.
 
+A cookie is also subject to certain size limits. Per [[RFC6265bis#section-5.5|Cookies: HTTP State Management Mechanism §Storage Model]]:
+* The combined lengths of the name and value fields must not be greater than 4096 [=byte|bytes=] (the <dfn for=cookie>maximum name/value pair size</dfn>).
+* The length of every field except the name and value fields must not be greater than 1024 [=bytes|bytes=] (the <dfn for=cookie>maximum attribute value size</dfn>).
+
 <!-- ============================================================ -->
 ## Cookie Store ## {#cookie-store--concept}
 <!-- ============================================================ -->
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1099,18 +1099,25 @@ optional |expires|,
 |sameSite|,
 run the following steps:
 
+1. If |name|, |value|, |domain|, or |path| contain U+003B (`;`), any [=C0 control=] character, or U+007F, then return failure.
 1. If |name|'s [=string/length=] is 0 and |value| contains U+003D (`=`), then return failure.
 1. If |name|'s [=string/length=] is 0 and |value|'s [=string/length=] is 0, then return failure.
+1. Determine whether |name| and |value| together exceed cookie size limits using these steps:
```
inexorabletash, line 1105: You don't need this as a step - it doesn't actually do anything in the algorithm. While it adds structure, it's more of a comment. Remove it?

index.bs
```diff
@@ -1099,18 +1099,25 @@ optional |expires|,
...
+    1. [=UTF-8 encode=] |name| and then compute its [=byte sequence=] [=byte sequence/length=].
```
inexorabletash, line 1106: It's unclear if the intent is to mutate _name_ here from being a string to a byte sequence, or to make an anonymous temporary. (Pretending this is code.)

Since _name_ and _value_ are used below with a sloppy (I wrote it!) invocation of encode, how about redoing this as something like:

1. Let _encodedName_ be the result of UTF-8 encoding _name_.
2. Let _encodedValue_ be the result of UTF-8 encoding _value_.
3. If the byte sequence length of encodedName plus the byte sequence length of encodedValue is greater than 4096, then return failure.

index.bs
```diff
@@ -1099,18 +1099,25 @@ optional |expires|,
...
+    1. [=UTF-8 encode=] |value| and then compute its [=byte sequence=] [=byte sequence/length=].
+    1. If the sum of the two computed lengths is greater than 4096 [=byte|bytes=], then return failure.
```
inexorabletash, line 1108: To avoid these magic numbers (4096 here, 1024 below), would it make sense to define "constants" for them, referencing the RFC?

index.bs
```diff
@@ -1099,18 +1099,25 @@ optional |expires|,
...
 1. Let |host| be |url|'s [=url/host=]
 1. Let |attributes| be a new [=/list=].
 1. If |domain| is not null, then run these steps:
+    1. If |domain|'s [=UTF-8 encode|UTF-8-encoded=] [=byte sequence=] [=byte sequence/length=] is greater than 1024 [=byte|bytes=], then return failure.
```
inexorabletash, line 1112: Maybe introduce _domainEncoded_ here too, and break this into two steps?

Also, maybe move this to be the last test/possible failure before appending?

index.bs
```diff
@@ -1099,18 +1099,25 @@ optional |expires|,
...
     1. If |domain| starts with U+002D (`.`), then return failure.
     1. If |host| does not equal |domain| and
         |host| does not end with U+002D (`.`) followed by |domain|,
         then return failure.
     1. [=list/Append=] \``Domain`\`/|domain| ([=encoded=]) to |attributes|.
 1. If |expires| is given, then [=list/append=] \``Expires`\`/|expires| ([=date serialized=]) to |attributes|.
 1. If |path| is not null, then run these steps:
+    1. If |path|'s [=UTF-8 encode|UTF-8-encoded] [=byte sequence=] [=byte sequence/length=] is greater than 1024 [=byte|bytes=], then return failure.
```
inexorabletash, line 1120: Same as above for _pathEncoded_ maybe?

index.bs
recvfrom: Thanks for the thorough feedback! 💯

For this one, are you thinking I should do something like:
```
1.  Let |namePlusValueMaxSize| be 4096 [=byte|bytes=] (as specified in [[RFC6265bis#section-5.5|RFC6265bis]]).
...
1.  If the sum of the two computed lengths is greater than |namePlusValueMaxSize| [=byte|bytes=], then return failure.
```

or is there a more elegant way to achieve this?

inexorabletash: For the constant, I'd do a <dfn> earlier in the spec... in the cookie-concept section which acts as a bridge to the RFC, maybe something like:

`The <dfn for=cookie>maximum name plus size</dfn> for a cookie is 4096, per Cookies: HTTP State blah blah blah`

and then in the algorithm:

`If the [=byte sequence=] [=byte sequence/length=] of |encodedName| plus the [=byte sequence=] [=byte sequence/length=] of encodedValue is greater than the [=cookie/maximum name plus size=], then return failure.`

recvfrom:
I believe I've addressed all the prior feedback - let me know if there's anything else that you think could be improved.

index.bs
```diff
@@ -1099,20 +1103,28 @@ optional |expires|,
 |sameSite|,
 run the following steps:
 
+1. If |name|, |value|, |domain|, or |path| contain U+003B (`;`), any [=C0 control=] character, or U+007F, then return failure.
 1. If |name|'s [=string/length=] is 0 and |value| contains U+003D (`=`), then return failure.
 1. If |name|'s [=string/length=] is 0 and |value|'s [=string/length=] is 0, then return failure.
+1. Let |encodedName| be the result of [=UTF-8 encode|UTF-8 encoding=] |name|.
+1. Let |encodedValue| be the result of [=UTF-8 encode|UTF-8 encoding=] |value|.
+1. If the [=byte sequence=] [=byte sequence/length=] of |encodedName| plus the [=byte sequence=] [=byte sequence/length=] of |encodedValue| is greater than the <a for=cookie>maximum name/value pair size</a>, then return failure.
```
inexorabletash, line 1111: Can you use the `[=cookie/maximum name/value pair size=]` syntax here for consistency?

inexorabletash (changes requested):
One nitpick, otherwise LGTM!

inexorabletash (changes requested):
Oops, one more thing:

In the places where it says e.g. "domain (encoded)", "path (encoded)", "name (encoded)" and "value (encoded)", can you change it to use domainEncoded (etc) ?

index.bs
recvfrom: I tried that initially but it looks like the slash breaks things:
```
      LINE ~1111: No 'dfn' refs found for 'value pair size'.
      WARNING: Unexported dfn that's not referenced locally - did you mean to export it?
      <dfn data-dfn-for="cookie" data-dfn-type="dfn" id="cookie-maximum-name-value-pair-size" data-lt="maximum name/value pair size" data-noexport="by-default">maximum name/value pair size<a href="#cookie-maximum-name-value-pair-size" class="self-link"></a></dfn>
```
should we use something else instead of `maximum name/value pair size`?

inexorabletash: Oh, oops! Stick with the <a> then.

recvfrom:
AH, I totally missed those. I understand now what you meant when you originally commented that name and value were used encoded in certain steps already. Anyway, I have replaced these instances.

Also, I realized that the RFC link I added was from a newer version of the draft than was linked to elsewhere, which could be a bit confusing.  I went ahead and updated all of the existing references to point to the latest draft, attempting to preserve the original intent behind the linking.  If you think that needs to be in a separate PR or if you think more changes are needed as a result of this, just let me know.  Thanks!

index.bs
```diff
@@ -992,7 +996,7 @@ The <dfn attribute for=ServiceWorkerGlobalScope>cookieStore</dfn> getter steps a
 
 [=Cookie=] attribute-values are stored as [=byte sequences=], not strings.
 
-To <dfn>encode</dfn> a |string|, run [=UTF-8 encode=] on |string|.
+To encode a |string|, run [=UTF-8 encode=] on |string|.
```
inexorabletash, line 999: Maybe we don't need this any more, if you're getting an unused definition warning?

inexorabletash (approved):
Left a comment, but let's merge/iterate.

index.bs
recvfrom: Indeed

```
WARNING: Unexported dfn that's not referenced locally - did you mean to export it?
      <dfn data-dfn-type="dfn" id="encode" data-lt="encode" data-noexport="by-default">encode<a href="#encode" class="self-link"></a></dfn>
```

PR #204, opened 2021-08-09
Allow tab characters in cookies

### LINKED ISSUE #203: Allow tab characters in cookies
Per https://github.com/httpwg/http-extensions/pull/1589, the tab character should no longer be treated as an invalid control character

recvfrom:
Fixes #203 - Allow tab characters in cookies

~~Also, applies the invalid character checks to all other attribute values as well for added clarity (even if they are less likely to contain invalid characters due to other restrictions on the types for those)~~

***
<a href="https://pr-preview.s3.amazonaws.com/recvfrom/cookie-store/pull/204.html" title="Last updated on Aug 13, 2021, 9:52 PM UTC (cd44859)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/204/2fba1a5...recvfrom:cd44859.html" title="Last updated on Aug 13, 2021, 9:52 PM UTC (cd44859)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1103,7 +1103,7 @@ optional |expires|,
 |sameSite|,
 run the following steps:
 
-1. If |name|, |value|, |domain|, or |path| contain U+003B (`;`), any [=C0 control=] character, or U+007F, then return failure.
+1. If |name| or |value| contain U+003B (`;`), any [=C0 control=] character except U+0009 (the horizontal tab character), or U+007F, then return failure.
 1. If |name|'s [=string/length=] is 0 and |value| contains U+003D (`=`), then return failure.
 1. If |name|'s [=string/length=] is 0 and |value|'s [=string/length=] is 0, then return failure.
 1. Let |encodedName| be the result of [=UTF-8 encode|UTF-8 encoding=] |name|.
--- a/index.bs
+++ b/index.bs
@@ -1104,6 +1104,9 @@ optional |expires|,
 run the following steps:
 
 1. If |name| or |value| contain U+003B (`;`), any [=C0 control=] character except U+0009 (the horizontal tab character), or U+007F, then return failure.
+
+    ISSUE(httpwg/http-extensions#1593): Note that it's up for discussion whether these character restrictions should also apply to |expires|, |domain|, |path|, and |sameSite| as well.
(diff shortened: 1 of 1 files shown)


recvfrom:
Hmm, after reading the latest version of the spec again, the changes in this PR don't exactly line up (likely due to an oversight)... I'll update this PR so that it aligns with what's in the spec today

inexorabletash:
Are there WPT changes that correspond to this?

recvfrom:
I just submitted the corresponding WPT changes for review - https://chromium-review.googlesource.com/c/chromium/src/+/3084521

index.bs
```diff
@@ -1103,7 +1103,7 @@ optional |expires|,
 |sameSite|,
 run the following steps:
 
-1. If |name|, |value|, |domain|, or |path| contain U+003B (`;`), any [=C0 control=] character, or U+007F, then return failure.
+1. If |name| or |value| contain U+003B (`;`), any [=C0 control=] character except U+0009 (the horizontal tab character), or U+007F, then return failure.
```
inexorabletash, line 1106: What happens to validation for domain and path?

recvfrom: I added a comment here about this - https://github.com/httpwg/http-extensions/issues/1593#issuecomment-895385796

In the RFC, the character restrictions apply to attribute values derived from Set-Cookie headers, but in the `Storage Model` section it only says that these character checks should happen for `name` and `value`...  I think this may be more of an oversight than intentional, but regardless, my thinking with the change above is that this spec should match what's in the RFC today and then when/if the character restrictions are made to apply to attribute values as well, we can update it here also. What do you think?

inexorabletash: SGTM, but I'd add a `TODO: ...` or `Issue: ...` block to the spec referencing the issue.

recvfrom: With what I added with the latest commit, it looks like this:

image

Does that seem reasonable?

inexorabletash: LGTM. Since the _expires_ and _sameSite_ attributes aren't passed through from script as strings I don't think those are a concern.

inexorabletash (approved):

PR #206, opened 2021-09-15
Add support for partitioned cookies to CookieStore.

DCtheTall:
***
<a href="https://pr-preview.s3.amazonaws.com/DCtheTall/cookie-store/pull/206.html" title="Last updated on Nov 3, 2021, 1:40 PM UTC (976425a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/206/e1a1cfa...DCtheTall:976425a.html" title="Last updated on Nov 3, 2021, 1:40 PM UTC (976425a)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/DCtheTall/cookie-store/pull/206.html" title="Last updated on Aug 4, 2023, 9:29 PM UTC (28de9d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/206/c9a8c1d...DCtheTall:28de9d0.html" title="Last updated on Aug 4, 2023, 9:29 PM UTC (28de9d0)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -369,7 +369,7 @@ await cookieStore.delete({
 ```
 
 If the site wants to delete the unpartitioned cookie, change the `partitioned`
-field to `false`.
+field to `false`. If the field is not present, the value defaults to `false`.
 
 ### Access all the cookie data
 
--- a/explainer.md
+++ b/explainer.md
@@ -14,9 +14,12 @@
   + [Read a cookie](#read-a-cookie)
   + [Read multiple cookies](#read-multiple-cookies)
   + [Read the cookies for a specific URL](#read-the-cookies-for-a-specific-url)
+  + [Read a partitioned cookie](#read-a-partitioned-cookie)
 * [The Modifications API](#the-modifications-api)
   + [Write a cookie](#write-a-cookie)
+  + [Write a partitioned cookie](#write-a-partitioned-cookie)
   + [Delete a cookie](#delete-a-cookie)
+  + [Delete a partitioned cookie](#delete-a-partitioned-cookie)
   + [Access all the cookie data](#access-all-the-cookie-data)
 * [The Change Events API](#the-change-events-api)
   + [Get change events in documents](#get-change-events-in-documents)
@@ -263,6 +266,22 @@ any URL under their scope.
 Documents can only obtain the cookies at their current URL. In other words,
 the only valid `url` value in Document contexts is the document's URL.
 
+### Read a partitioned cookie
+
+If the user agent supports
+[cookie partitioning](https://github.com/DCtheTall/CHIPS), then the cookie
+objects will have a boolean value indicating if the cookie is partitioned.
+
+```javascript
+// Read a cookie set without the Partitioned attribute.
+const cookie = await cookieStore.get('session_id');
+console.log(cookie.partitioned);  // -> false
+
+// Read a Partitioned cookie from a third-party context.
+const cookie = await cookieStore.get('__Host-third_party_session_id');
+console.log(cookie.partitioned);  // -> true
+```
+
 ## The Modifications API
 
 Both documents and service workers access the same modification API, via the
@@ -292,6 +311,21 @@ await cookieStore.set({
 });
 ```
 
+### Write a partitioned cookie
+
+If the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS)
+then you can set a partitioned cookie in a third-party context using the following.
+
+```javascript
+await cookieStore.set({
+  name: '__Host-third_party_session_id',
+  value: 'foobar',
+  path: '/',
+  sameSite: 'none'
+  // `Secure` is implicitly set
+});
+```
+
 ### Delete a cookie
 
 ```javascript
@@ -315,6 +349,27 @@ try {
   console.error(`Failed to delete cookie: ${e}`);
 }
 ```
+
+### Delete a partitioned cookie
+
+If the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS)
+then it is possible for a site to set both a partitioned and unpartitioned
+cookie with the same name.
+
+In this edge case, if a site would like to distinguish between whether they
+want to delete their partitioned and unpartitioned cookie, they can provide
+a `partitioned` attribute. If the site wants to delete the partitioned cookie,
+the site could use:
+
+```javascript
+await cookieStore.delete({
+  name: '__Host-third_party_session_id',
+  partitioned: true
+});
+```
+
+If the site wants to delete the unpartitioned cookie, change the `partitioned`
+field to `false`.
 
 ### Access all the cookie data
 
--- a/explainer.md
+++ b/explainer.md
@@ -278,7 +278,10 @@ const cookie = await cookieStore.get('session_id');
 console.log(cookie.partitioned);  // -> false
 
 // Read a Partitioned cookie from a third-party context.
-const cookie = await cookieStore.get('__Host-third_party_session_id');
+const cookie = await cookieStore.get({
+  name: '__Host-third_party_session_id',
+  partitioned: true
+});
 console.log(cookie.partitioned);  // -> true
 ```
 
@@ -321,7 +324,8 @@ await cookieStore.set({
   name: '__Host-third_party_session_id',
   value: 'foobar',
   path: '/',
-  sameSite: 'none'
+  sameSite: 'none',
+  partitioned: true
   // `Secure` is implicitly set
 });
 ```
--- a/explainer.md
+++ b/explainer.md
@@ -268,9 +268,7 @@ the only valid `url` value in Document contexts is the document's URL.
 
 ### Read a partitioned cookie
 
-If the user agent supports
-[cookie partitioning](https://github.com/DCtheTall/CHIPS), then the cookie
-objects will have a boolean value indicating if the cookie is partitioned.
+The cookie objects will have a boolean value indicating if the cookie is partitioned.
 
 ```javascript
 // Read a cookie set without the Partitioned attribute.
@@ -360,10 +358,7 @@ If the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS)
 then it is possible for a site to set both a partitioned and unpartitioned
 cookie with the same name.
 
-In this edge case, if a site would like to distinguish between whether they
-want to delete their partitioned and unpartitioned cookie, they can provide
-a `partitioned` attribute. If the site wants to delete the partitioned cookie,
-the site could use:
+To delete a partitioned cookie, the `partitioned` parameter must be provided:
 
 ```javascript
 await cookieStore.delete({
@@ -372,8 +367,8 @@ await cookieStore.delete({
 });
 ```
 
-If the site wants to delete the unpartitioned cookie, change the `partitioned`
-field to `false`. If the field is not present, the value defaults to `false`.
+If the site wants to only delete the unpartitioned cookie, change the `partitioned`
+field to `false` or omit the property.  
 
 ### Access all the cookie data
 
--- a/index.bs
+++ b/index.bs
@@ -500,12 +500,14 @@ dictionary CookieInit {
   USVString? domain = null;
   USVString path = "/";
   CookieSameSite sameSite = "strict";
+  boolean partitioned = false;
 };
 
 dictionary CookieStoreDeleteOptions {
   required USVString name;
   USVString? domain = null;
   USVString path = "/";
+  boolean? partitioned;
 };
 
 dictionary CookieListItem {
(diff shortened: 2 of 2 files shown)


DCtheTall:
This PR is a change to the CookieStore API spec to support partitioned cookies.

These changes could be used to support the [Partitioned attribute, a.k.a. CHIPS](https://github.com/WICG/CHIPS).

There is an [open PR](https://github.com/WICG/CHIPS/pull/24) in the CHIPS repository adding a note about these changes to the explainer.

index.bs
```diff
@@ -500,12 +500,14 @@ dictionary CookieInit {
...
   USVString path = "/";
   CookieSameSite sameSite = "strict";
+  boolean partitioned = false;
 };
 
 dictionary CookieStoreDeleteOptions {
   required USVString name;
   USVString? domain = null;
   USVString path = "/";
+  boolean? partitioned;
```
domenic, line 510:
```suggestion
  boolean partitioned;
```

DCtheTall: Done.

DCtheTall:
One quick addendum: if user agents implement service worker partitioning ([ref](https://github.com/privacycg/storage-partitioning)) then the cookie's partition key should be the site of the worker's partition key.

If user agents don't implement storage partitioning, then partitioned cookies may need to be blocked in service workers. Otherwise, service workers could use partitioned cookies as a cross-site identifier. See [the CHIPS explainer](https://github.com/WICG/CHIPS#service-workers) for more detail.

DCtheTall:
I put together a [WIP CL](https://chromium-review.googlesource.com/c/chromium/src/+/3212611) that implements these changes in Chromium.

The CL fails some web platform tests (for now 🤞), but you can get the general idea of how these changes would materialize from it.

inexorabletash:
Wow, this fell off my radar. Is this just waiting for final review/merge, or is there further discussion that should happen first?

index.bs
```diff
@@ -689,6 +692,9 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
         |options|["{{CookieInit/domain}}"],
         |options|["{{CookieInit/path}}"], and
         |options|["{{CookieInit/sameSite}}"].
+    1. If the |options|["{{CookieInit/partitioned}}"] attribute is present and the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS), then
+      1. If the current browsing context is a service worker, set the cookie's partition key be |origin|'s [=site=].
```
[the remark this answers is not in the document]
inexorabletash: ... and "current browsing context" isn't the right term. I think "current global object" ?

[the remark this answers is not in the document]
inexorabletash: Also.. this is run after the cookie is set, which seems unusual. ISTM the partition key should be captured before "set a cookie" and passed into it.

explainer.md
```diff
@@ -316,6 +354,27 @@ try {
...
+
+```javascript
+await cookieStore.delete({
+  name: '__Host-third_party_session_id',
+  partitioned: true
+});
+```
+
+If the site wants to delete the unpartitioned cookie, change the `partitioned`
+field to `false`. If the field is not present, the value defaults to `false`.
```
inexorabletash, line 376: In the IDL the default isn't given for `CookieStoreDeleteOptions`

index.bs
```diff
@@ -743,6 +749,11 @@ The <dfn method for=CookieStore>delete(|options|)</dfn> method steps are:
         |options|["{{CookieStoreDeleteOptions/name}}"],
         |options|["{{CookieStoreDeleteOptions/domain}}"], and
         |options|["{{CookieStoreDeleteOptions/path}}"].
+    1. If the |options|["{{CookieStoreDeleteOptions/partitioned}}"] field is present and the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS), then
```
inexorabletash, line 752: "field" isn't the right terminology.  Also, per previous comments, if we can default this it simplifies the logic.

explainer.md
```diff
@@ -263,6 +266,22 @@ any URL under their scope.
...
+If the user agent supports
+[cookie partitioning](https://github.com/DCtheTall/CHIPS), then the cookie
+objects will have a boolean value indicating if the cookie is partitioned.
+
+```javascript
+// Read a cookie set without the Partitioned attribute.
+const cookie = await cookieStore.get('session_id');
+console.log(cookie.partitioned);  // -> false
+
+// Read a Partitioned cookie from a third-party context.
```
dmdabbs, line 280: CHIPS [does not mandate](https://github.com/privacycg/CHIPS#handling-older-or-incompatible-clients) use of the __Host-prefix, 

>Although it is not required, it is still recommended to still include the __Host- prefix. Even clients that do not recognize the Partitioned attribute still enforce the semantics of the __Host- prefix. 

so a request could present two cookies with the same name. Will `cookieStore.get()` fail or simply return the first of the two cookies? In that situation, would `cookieStore.getAll()` retrieve the two cookies?

inexorabletash: #221 is relevant here - there are existing situations where multiple cookies with the same name can be present, not a new situation with partitioned cookies.

Currently as spec'd and implemented in Chrome, the first cookie is returned.

explainer.md
```diff
@@ -292,6 +314,22 @@ await cookieStore.set({
 });
 ```
 
+### Write a partitioned cookie
+
+If the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS)
```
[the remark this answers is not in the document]
DCtheTall: Done

explainer.md
```diff
@@ -316,6 +354,27 @@ try {
 }
 ```
 
+### Delete a partitioned cookie
+
+If the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS)
```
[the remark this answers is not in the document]
DCtheTall: Done

explainer.md
```diff
@@ -316,6 +354,27 @@ try {
...
+then it is possible for a site to set both a partitioned and unpartitioned
+cookie with the same name.
+
+In this edge case, if a site would like to distinguish between whether they
```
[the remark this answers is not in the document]
DCtheTall: Changed the wording

index.bs
```diff
@@ -689,6 +692,9 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
         |options|["{{CookieInit/domain}}"],
         |options|["{{CookieInit/path}}"], and
         |options|["{{CookieInit/sameSite}}"].
+    1. If the |options|["{{CookieInit/partitioned}}"] attribute is present and the user agent supports [cookie partitioning](https://github.com/WICG/CHIPS), then
+      1. If the current browsing context is a service worker, set the cookie's partition key be |origin|'s [=site=].
+      1. If the current browsing context is an HTML document, then the cookie's partition key is the [=site=] of the topmost frame.
```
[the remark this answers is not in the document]
DCtheTall: Removed mentioning of partition key in this section.

explainer.md
DCtheTall: Added default to IDL

index.bs
DCtheTall: Done

explainer.md
```diff
@@ -263,6 +266,25 @@ any URL under their scope.
 Documents can only obtain the cookies at their current URL. In other words,
 the only valid `url` value in Document contexts is the document's URL.
 
+### Read a partitioned cookie
+
+If the user agent supports
+[cookie partitioning](https://github.com/DCtheTall/CHIPS), then the cookie
```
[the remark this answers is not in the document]
DCtheTall: Done, forgot to resolve.

index.bs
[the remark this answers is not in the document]
DCtheTall: I think the partition key can be up to RFC6265bis to define, since the CookieStore API is not actually concerned with partition key values and the cookie RFC will (some day) specify how to select which partitioned cookies to include in a request. So I do not think this spec needs to be concerned with that?

index.bs
```diff
@@ -764,6 +770,11 @@ The <dfn method for=CookieStore>delete(|options|)</dfn> method steps are:
         |options|["{{CookieStoreDeleteOptions/name}}"],
         |options|["{{CookieStoreDeleteOptions/domain}}"], and
         |options|["{{CookieStoreDeleteOptions/path}}"].
+    1. If |options|["{{CookieStoreDeleteOptions/partitioned}}"] is present and the user agent supports [cookie partitioning](https://github.com/privacycg/CHIPS), then
```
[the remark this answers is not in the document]
DCtheTall: Done.

index.bs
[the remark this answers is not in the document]
inexorabletash: Yes... but https://wicg.github.io/cookie-store/#cookie-concept basically has "proxy" definitions for RFC6265bis terms, since I don't think we can link directly to them. We can add it later though.

index.bs
```diff
@@ -698,6 +701,7 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
 1. Let |origin| be |settings|'s [=environment settings object/origin=].
 1. If |origin| is an [=opaque origin=], then return [=a promise rejected with=] a "{{SecurityError}}" {{DOMException}}.
 1. Let |url| be |settings|'s [=environment/creation URL=].
+1. Let |partitioned| be true if |options|["{{CookieInit/partitioned}}"] is present and true, or false otherwise.
```
inexorabletash, line 704: Since this defaults to true now, this step can be removed...

index.bs
```diff
@@ -706,8 +710,9 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
         |options|["{{CookieInit/value}}"],
         |options|["{{CookieInit/expires}}"],
         |options|["{{CookieInit/domain}}"],
-        |options|["{{CookieInit/path}}"], and
-        |options|["{{CookieInit/sameSite}}"].
+        |options|["{{CookieInit/path}}"],
+        |options|["{{CookieInit/sameSite}}"], and
+        |partitioned|.
```
inexorabletash, line 715: ... and this can just be `|options|["{{CookieStoreDeleteOptions/partitioned}}"]` now.

index.bs
```diff
@@ -762,8 +767,9 @@ The <dfn method for=CookieStore>delete(|options|)</dfn> method steps are:
     1. Let |r| be the result of running [=delete a cookie=] with
         |url|,
         |options|["{{CookieStoreDeleteOptions/name}}"],
-        |options|["{{CookieStoreDeleteOptions/domain}}"], and
-        |options|["{{CookieStoreDeleteOptions/path}}"].
+        |options|["{{CookieStoreDeleteOptions/domain}}"],
+        |options|["{{CookieStoreDeleteOptions/path}}"],
+        |options|["{{CookieStoreDeleteOptions/partitioned}}"]
```
inexorabletash, line 772: nit: the sentence lost the "and" and the trailing period.

inexorabletash (changes requested):
Almost there! Just tiny nitpicks.

inexorabletash:
BTW, I was playing with cookieStore/document.cookie interop and it took me a while to realize that with document.cookie "partitioned" is ignored unless "path" and "secure" are also specified, which are the default with cookieStore but of course not document.cookie. Do you think we should add an example in the "Modifying Cookies" section, or is this well known?

index.bs
DCtheTall: Done.

index.bs
DCtheTall: Done.

index.bs
DCtheTall: Done.

DCtheTall:
> BTW, I was playing with cookieStore/document.cookie interop and it took me a while to realize that with document.cookie "partitioned" is ignored unless "path" and "secure" are also specified

Tad correction, you only need the "secure" attribute set. The "path" attribute is not necessary.

> Do you think we should add an example in the "Modifying Cookies" section, or is this well known?

The Secure requirement is a well-known part of the proposal for the definition of Partitioned we have for RFC6265bis.

inexorabletash:
> > BTW, I was playing with cookieStore/document.cookie interop and it took me a while to realize that with document.cookie "partitioned" is ignored unless "path" and "secure" are also specified
> 
> Tad correction, you only need the "secure" attribute set. The "path" attribute is not necessary 😄

Thanks for clarifying! https://source.chromium.org/chromium/chromium/src/+/main:net/cookies/canonical_cookie_unittest.cc;l=731 mislead me a bit I guess.

DCtheTall:
> Thanks for clarifying! [source.chromium.org/chromium/chromium/src/+/main:net/cookies/canonical_cookie_unittest.cc;l=731](https://source.chromium.org/chromium/chromium/src/+/main:net/cookies/canonical_cookie_unittest.cc;l=731) mislead me a bit I guess.

Ah, this is because Path=/ is required for the __Host- prefix, which that cookie also uses. I can see why that is initially confusing, especially considering Partitioned was originally designed to require __Host- but we loosened it to only require Secure 😄

PR #214, opened 2023-02-15
Normative: Change use of EpochTimeStamp to DOMHighResTimeStamp

### LINKED ISSUE #205: Consider using DOMHighResTimestamp instead of DOMTimeStamp
We're considering our options RE `DOMTimeStamp`, and wondering what it's used for. It seems less well-defined than `DOMHighResTimestamp`. Would y'all consider switching over?

See heycam/webidl#2 for discussion.

inexorabletash:
EpochTimeStamp is a typedef for unsigned long long and deprecated, and DOMHighResTimeStamp is a typedef for double and the preferred type going forward. Semantics are the same.

Resolves #205

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/214.html" title="Last updated on Feb 18, 2023, 12:57 AM UTC (af6c826)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/214/5dfe3db...af6c826.html" title="Last updated on Feb 18, 2023, 12:57 AM UTC (af6c826)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -506,7 +506,7 @@ enum CookieSameSite {
 dictionary CookieInit {
   required USVString name;
   required USVString value;
-  EpochTimeStamp? expires = null;
+  DOMHighResTimeStamp? expires = null;
   USVString? domain = null;
   USVString path = "/";
   CookieSameSite sameSite = "strict";
@@ -523,7 +523,7 @@ dictionary CookieListItem {
   USVString value;
   USVString? domain;
   USVString path;
-  EpochTimeStamp? expires;
+  DOMHighResTimeStamp? expires;
   boolean secure;
   CookieSameSite sameSite;
 };
@@ -1021,7 +1021,7 @@ Note: This is the same representation used for [=time values=] in [[ECMAScript]]
 
 
 <div algorithm>
-To <dfn>date serialize</dfn> a {{EpochTimeStamp}} |millis|,
+To <dfn>date serialize</dfn> a {{DOMHighResTimeStamp}} |millis|,
 let |dateTime| be the date and time |millis| milliseconds after 00:00:00 UTC, 1 January 1970
 (assuming that there are exactly 86,400,000 milliseconds per day),
 and return a [=byte sequence=] corresponding to the closest `cookie-date` representation of |dateTime| according to [[RFC6265bis#section-5.1.1]]<!-- Dates -->.


ayuishii (approved):
LGTM, thanks for the quick update!

domenic:
Semantics of the two are *not* the same: DOMHighResTimeStamp is time since document creation; EpochTimeStamp is time since the epoch.

domenic:
Filed https://github.com/w3c/hr-time/issues/149

ayuishii:
Thanks @domenic for the clarification, and filing the issue!

inexorabletash:
Hmmm... _"The DOMHighResTimeStamp type is used to store a duration in milliseconds. Depending on its context, it may represent the moment that is this duration after a base moment like a time origin **or the Unix epoch.**"_  (emphasis mine)

The algorithms https://wicg.github.io/cookie-store/#as-a-timestamp and https://wicg.github.io/cookie-store/#date-serialize are explicit about the base moment and specify ignoring leap seconds, so at least from cookie store's perspective they're the same I think?

But thanks for jumping in @domenic ! Your guidance is always welcome.

domenic:
Oh, I might just be wrong then... I didn't recall any instances of DOMHighResTimeStamp being epoch-relative, but looking through [Webdex](https://dontcallmedom.github.io/webdex/d.html#DOMHighResTimeStamp%40%40%40%40typedef) I found that [WebTransport is using it that way](https://w3c.github.io/webtransport/). So this is probably fine to proceed, although I still find the naming of the two types confusing and will discuss that in the issue I opened.

inexorabletash:
We're not in a rush to land this (also blocked on Bikeshed etc fun) so I'm happy to wait for the other discussion to make progress.

domenic:
My tentative conclusion (https://github.com/w3c/hr-time/issues/149#issuecomment-1432299178) is that landing this, and doing the same thing to the Notifications API, would be a good thing because then we could kill `EpochTimeStamp` entirely and I would be less confused. But indeed, hearing from a few other folks (like the HR-TIME editors) would be welcome.

inexorabletash:
Going ahead and merging this. We haven't figured out a way to test it from script. Also I revisited the hr-time definitions and I don't see our conversions [to](https://wicg.github.io/cookie-store/#as-a-timestamp) or [from](https://wicg.github.io/cookie-store/#date-serialize) a timestamp being simplified (although we could align wording slightly better).

PR #217, opened 2023-07-07
Normative: get()'s promise should resolve to null, not undefined

### LINKED ISSUE #210: Change get() to return null instead of undefined
When get() was implemented in Chrome, we failed to match our own spec and made get() return null instead of undefined if no matching cookie was found. The WPT tests match the implementation, not the spec. Oops.

Someone noticed and filed https://bugs.chromium.org/p/chromium/issues/detail?id=1355228

@ayuishii investigated HTTP Archive and found several sites/libraries that seem to do strict comparisons against null. 

Some examples from the the first 1000 mobile results:

* These ones look safe with either null or undefined:
  * controltag.js (a common library - salesforce?) contains `return t=this.cookieStore.get(n),(null!=t?t.length:void 0)||(...` - this is a sloppy (not strict) comparison, and since `undefined == null` this code would not break (491/1000)
  * "modern.js" contains `var t=this.cookieStore.get(r);return null==t?null:"true"===t` - also fine. (284/1000)
  * a library used by many sites contains `if ($cookieStore.get('moyasarId') === null || $cookieStore.get('moyasarId') === '' || $cookieStore.get('moyasarId') === undefined) {` - this seems like it would handle null and undefined equally (23/1000)
* These look like they'd fail if we changed the return value:
  * library served off of wovn.io: `var n=this.cookieStore.get(a.WovnCookies.WOVN_MTM_SHOWED_LANGS);return null!==n?JSON.parse(n):[]` (130/1000)

Notably, `JSON.parse(undefined)` throws while `JSON.parse(null)` returns `null`, and JSON-parsing cookies seems to be fairly common (a handful of other pages, e.g. https://www.justfab.se/assets/dist/js/vueapp_m-c5291be04a.js has `t=this.cookieStore.get(this.$options.cookie.name),t=JSON.parse(t)`

So... I suspect it is not web-compatible to fix and we should instead update the spec to have get() return null. _sigh_

Thoughts?

inexorabletash:
The spec and tests were inconsistent:

* In the spec's IDL, the inner type of the promise returned by get() is nullable.

* In the spec's prose, the steps call for resolving the promise with undefined.

* In WPT, the tests[1] assert that the promise is resolved with null.

The Chromium implementation[2] matched the tests, not the spec prose.

Investigation documented in #210 indicates it is likely not web-compatible to make this change purely for idealistic reasons.

Resolves #210.

[1] example: https://github.com/web-platform-tests/wpt/blob/master/cookie-store/cookieStore_delete_arguments.https.any.js

[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/cookie_store/cookie_store.cc;l=482

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/217.html" title="Last updated on Jul 7, 2023, 8:50 PM UTC (2641c90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/217/837f4f2...2641c90.html" title="Last updated on Jul 7, 2023, 8:50 PM UTC (2641c90)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -559,7 +559,7 @@ The <dfn method for=CookieStore>get(|name|)</dfn> method steps are:
     1. Let |list| be the results of running [=query cookies=] with
         |url| and |name|.
     1. If |list| is failure, then [=/reject=] |p| with a {{TypeError}} and abort these steps.
-    1. If |list| [=list/is empty=], then [=/resolve=] |p| with undefined.
+    1. If |list| [=list/is empty=], then [=/resolve=] |p| with null.
     1. Otherwise, [=/resolve=] |p| with the first item of |list|.
 1. Return |p|.
 
@@ -586,7 +586,7 @@ The <dfn method for=CookieStore>get(|options|)</dfn> method steps are:
         |url| and
         |options|["{{CookieStoreGetOptions/name}}"] (if present).
     1. If |list| is failure, then [=reject=] |p| with a {{TypeError}} and abort these steps.
-    1. If |list| [=list/is empty=], then [=/resolve=] |p| with undefined.
+    1. If |list| [=list/is empty=], then [=/resolve=] |p| with null.
     1. Otherwise, [=/resolve=] |p| with the first item of |list|.
 1. Return |p|.
 


inexorabletash:
@ayuishii & @annevk - this look okay?

annevk (approved):

ayuishii (approved):
LGTM, thanks!

Issue #189, opened 2020-12-02
Deal with non-UTF-8 cookies


annevk:
I think I raised this before, but I don't really see text about it in the specification. If you limit yourself to UTF-8, you need to deal with cookies that do not decode as UTF-8, presumably by dropping them as Chrome appears to do for `document.cookie`. That should be specified somehow.

I think that means cookies need to use "UTF-8 decode without BOM or fail" from Encoding.

annevk:
Ideally the way this would be specified is by sharing logic with `document.cookie`, FWIW. (Something I also raised before.)

inexorabletash:
Thanks and +1. @bsittler did substantial investigation into cross-browser and API/header behavior with encoding, but I'm not sure that research was retained and was definitely not reflected into the spec.

We should audit test coverage and get coverage/spec updates.

(Sharing w/ HTML also SGTM. PRs/concrete suggestions welcome!)

inexorabletash:
FYI, here's the current behavior in Chrome:

```js
cookie_test(async t => {
  const value = 'value\u00C6'; // Æ
  const utf8_value = 'value\xC3\x86'; // UTF-8 encoding
  const non_utf8_value = 'value\xC6'; // ISO-8859-1 encoding

  await setCookieBinaryHttp(`utf8=${utf8_value}; path=/`);
  assert_equals(await getCookieBinaryHttp(), `utf8=${utf8_value}`);

  // Set a non-UTF-8 cookie on the server, verify the server thinks its there.
  await setCookieBinaryHttp(`nonutf8=${non_utf8_value}; path=/`);
  assert_equals(await getCookieBinaryHttp(), `utf8=${utf8_value}; nonutf8=${non_utf8_value}`);

  // document.cookie vs. mixed cookie encodings
  assert_equals(await getCookieStringDocument(), undefined,
                'document.cookie yields nothing if non-UTF-8 cookies served');

  // Cookie Store API vs. mixed cookie encodings
  const cookies = await cookieStore.getAll();
  assert_equals(cookies.length, 2, 'Both cookies are present by name');
  assert_not_equals(
      cookies.find(c => c.name === 'utf8'), undefined, 'utf8 cookie present');
  assert_equals(
      cookies.find(c => c.name === 'utf8').value, value,
      'utf8 cookie value should be preserved');
  assert_not_equals(
      cookies.find(c => c.name === 'nonutf8'), undefined,
      'nonutf8 cookie present');
  assert_equals(
[...]

PR #218, opened 2023-07-07
Fix up encoding definitions/references

### LINKED ISSUE #202: Fix up encoding definitions/references
Following dd2ddc7ce8b081ef10d913fc71b862059d8c1a4f there's this text:

> To encode a string, run UTF-8 encode on string.

This text is unnecessary, as places that need to encode invoke "UTF-8 encode" directly. We should either:

* Remove that line; or
* Define **encode** and then simplify the call sites from _"result of UTF-8 encoding var"_ to _"result of encoding var"_

At the same time, we should consider if this definition is worthwhile:

> To **decode** a value, run UTF-8 decode without BOM on value.

In theory we could remove that and make the UTF-8 decoding inline, but the "without BOM" part makes it wordier. 

cc: @recvfrom

inexorabletash:
Following dd2ddc7 a definition for "encode" was unneeded - relevant places invoked "UTF-8 encode" directly. This is removed.

Based on guidance from @domenic, having a specific "decode" definition made the spec harder to read; replace it with direct references to running "UTF-8 decode without BOM".

Finally, turn an orphaned paragraph into a NOTE about cookie attributes being byte sequences, since it was wortk keeping.

(Also, silence a Bikeshed warning by replacing |variable| references in domintro sections with <var ignore>.)

Resolves #202

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/pull/218.html" title="Last updated on Jul 7, 2023, 10:39 PM UTC (75fb6dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/218/962cfd2...75fb6dd.html" title="Last updated on Jul 7, 2023, 10:39 PM UTC (75fb6dd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -444,6 +444,8 @@ A cookie is also subject to certain size limits. Per [[RFC6265bis#section-5.6]]<
 * The combined lengths of the name and value fields must not be greater than 4096 [=bytes=] (the <dfn for=cookie>maximum name/value pair size</dfn>).
 * The length of every field except the name and value fields must not be greater than 1024 [=bytes=] (the <dfn for=cookie>maximum attribute value size</dfn>).
 
+NOTE: [=Cookie=] attribute-values are stored as [=byte sequences=], not strings.
+
 <!-- ============================================================ -->
 ## Cookie Store ## {#cookie-store--concept}
 <!-- ============================================================ -->
@@ -537,8 +539,8 @@ typedef sequence<CookieListItem> CookieList;
 <!-- ============================================================ -->
 
 <div class="domintro note">
-    : |cookie| = await cookieStore . {{CookieStore/get(name)|get}}(|name|)
-    : |cookie| = await cookieStore . {{CookieStore/get(options)|get}}(|options|)
+    : |cookie| = await cookieStore . {{CookieStore/get(name)|get}}(<var ignore>name</var>)
+    : |cookie| = await cookieStore . {{CookieStore/get(options)|get}}(<var ignore>options</var>)
 
     ::  Returns a promise resolving to the first in-scope [=script-visible=] value
         for a given cookie name (or other options).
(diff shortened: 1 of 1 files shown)


inexorabletash:
@ayuishii and @annevk can you take a look?

The text here would change when #189 is tackled, i.e. we'd reference "...or fail" and define failure behavior.

index.bs
```diff
@@ -1073,10 +1069,10 @@ run the following steps:
...
 To <dfn>create a {{CookieListItem}}</dfn> from |cookie|, run the following steps.
 
-1. Let |name| be |cookie|'s [=cookie/name=] ([=decoded=]).
-1. Let |value| be |cookie|'s [=cookie/value=] ([=decoded=]).
-1. Let |domain| be |cookie|'s [=cookie/domain=] ([=decoded=]).
-1. Let |path| be |cookie|'s [=cookie/path=] ([=decoded=]).
+1. Let |name| be the result of running [=UTF-8 decode without BOM=] on |cookie|'s [=cookie/name=].
+1. Let |value| be the result of running [=UTF-8 decode without BOM=] on |cookie|'s [=cookie/value=].
+1. Let |domain| be the result of running [=UTF-8 decode without BOM=] on |cookie|'s [=cookie/domain=].
+1. Let |path| be the result of running [=UTF-8 decode without BOM=] on |cookie|'s [=cookie/path=].
```
annevk, line 1075: Can these two contain non-ASCII?

annevk (approved):
This seems strictly editorial and looks good to me.

index.bs
inexorabletash: Per https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-syntax looks like everything is a SHOULD NOT not a MUST NOT. Given that:

* Path looks constrained to any U+0020...U+007E except `;`.  So ASCII only.
* Domain similarly looks constrained to ASCII given the notes. 

I guess we could use `ByteString` here?

annevk: Yeah, though you'd need additional validation on input either way.

ayuishii (approved):
Thanks for the change!

PR #219, opened 2023-07-07
Add checks for not potentially trustworthy and "file" origins.

### LINKED ISSUE #193: Add references to secure/insecure contexts/origins and clarify the case of local files.
This is mentioned at several places, but AFAIK the only reference is on the IDL: 

https://heycam.github.io/webidl/#SecureContext

which in turn links to

https://html.spec.whatwg.org/multipage/webappapis.html#secure-context

and in turn to

https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url
https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy

This seems to mean we should treat localhost, loopback and files:// addresses as secure when deciding whether a cookie can be modified. For consistency with `document.cookie`, it seems the chromium people want to allow cookie store writing API on localhost/loopback but not `files://` (see https://groups.google.com/a/chromium.org/g/blink-dev/c/ekdeaj09c0w/m/HVpvpMmHBQAJ ).

So I think:

1. The spec could more explicitly reference the secure contexts spec in the various places "secure" or "insecure" is used.
2. The spec should explicitly exclude `file://` for consistency with `document.cookie`.

Probably 2. can be done in a local definition based on https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url ; and 1. can link to that new definition.

inexorabletash:
This aligns the spec with Chromium's behavior, namely that writes where the origin is not potentially trustworthy or is "file" scheme result in failure with a TypeError.

Resolves #193

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -676,6 +676,7 @@ The <dfn method for=CookieStore>set(|name|, |value|)</dfn> method steps are:
 1. Let |settings| be [=/this=]'s [=/relevant settings object=].
 1. Let |origin| be |settings|'s [=environment settings object/origin=].
 1. If |origin| is an [=opaque origin=], then return [=a promise rejected with=] a "{{SecurityError}}" {{DOMException}}.
+1. If |origin| is not a [=/potentially trustworthy origin=], or if |origin|'s [=origin/scheme=] is "`file`", then return [=a promise rejected with=] a {{TypeError}}.
 1. Let |url| be |settings|'s [=environment/creation URL=].
 1. Let |p| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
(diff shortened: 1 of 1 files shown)


inexorabletash:
This PR (as currently written) is purely to align the spec text w/ Chromium behavior. That doesn't mean we should merge it though! Notably:

* Chromium doesn't error on read. Should it?
* `document.cookie` prevents writes if the document is "cookie-averse". Alignment might be nice?
* Needs tests!

ayuishii (approved):
Thanks for the change!
I think this LGTM.

inexorabletash:
I looked briefly at tests just to capture Chrome's behavior - given the `[SecureContext]` requirement for the API I'm drawing a blank on exercising the "not potentially trustworthy" check from WPT given https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy .

A manual test for `file:` is doable. Ideas welcome.

bakulf:
Currently, in Gecko, the CookieStore API is not available on potentially untrustworthy origins or pages loaded using the "file" scheme.

PR #224, opened 2024-02-05
Add '[SecureContext]` tags to the interfaces

OnkarRuikar:
- addresses https://github.com/w3c/webref/issues/1142#issuecomment-1924200755

`w3c/webref` repo automatically extracts syntaxes from these spec docs. At the moment some syntax sections are missing the `[SecureContext]` tags so it is [missing from extracted data in webref as well](https://github.com/w3c/webref/blob/1ebc07b4638f130623f054e556da62fd6a045e01/ed/idl/cookie-store.idl#L92).

The feature has been [marked secure in MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/ExtendableCookieChangeEvent). 

The PR adds the tags to the remaining interfaces.

***
<a href="https://pr-preview.s3.amazonaws.com/OnkarRuikar/cookie-store/pull/224.html" title="Last updated on Feb 5, 2024, 2:58 PM UTC (8044570)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/224/880b5ea...OnkarRuikar:8044570.html" title="Last updated on Feb 5, 2024, 2:58 PM UTC (8044570)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -895,7 +895,8 @@ The <dfn method for=CookieStoreManager>unsubscribe(|subscriptions|)</dfn> method
 The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}} via {{ServiceWorkerRegistration/cookies}} which provides the interface for subscribing to cookie changes.
 
 <xmp class=idl>
-[Exposed=(ServiceWorker,Window)]
+[Exposed=(ServiceWorker,Window),
+ SecureContext]
 partial interface ServiceWorkerRegistration {
   [SameObject] readonly attribute CookieStoreManager cookies;
 };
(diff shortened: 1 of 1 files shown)


annevk:
I think these end up being redundant because the interfaces they are extending are already limited to secure contexts. We usually don't specify things redundantly.

OnkarRuikar:
> I think these end up being redundant because the interfaces they are extending are already limited to secure contexts. We usually don't specify things redundantly.

I see. What about `ExtendableCookieChangeEvent`

index.bs
```diff
@@ -895,7 +895,8 @@ The <dfn method for=CookieStoreManager>unsubscribe(|subscriptions|)</dfn> method
 The {{ServiceWorkerRegistration}} interface is extended to give access to a {{CookieStoreManager}} via {{ServiceWorkerRegistration/cookies}} which provides the interface for subscribing to cookie changes.
 
 <xmp class=idl>
-[Exposed=(ServiceWorker,Window)]
+[Exposed=(ServiceWorker,Window),
+ SecureContext]
```
OnkarRuikar, line 899:
```suggestion
[Exposed=(ServiceWorker,Window)]
```

annevk:
If you're only exposed to the ServiceWorker global you are by definition limited to secure contexts because the ServiceWorker global is only exposed to secure contexts.

OnkarRuikar:
> If you're only exposed to the ServiceWorker global you are by definition limited to secure contexts because the ServiceWorker global is only exposed to secure contexts.

@annevk  We want to make it explicit so that the automated [`w3c/webref`](https://github.com/w3c/webref) extractor can pick this fact correctly.

annevk:
I don't think we want to add redundancy to standards for the benefit of tooling. Instead the tooling should be fixed. Could you file an issue against w3c/webref? Feel free to at-mention me if you think that would help.

Issue #252, opened 2025-03-13
Impossible to delete nameless cookie


valenting:
The [delete algorithm](https://wicg.github.io/cookie-store/#delete-cookie-algorithm) says:
> Let expires be the earliest representable date represented
> Let value be the empty string.
> Return the results of running [set a cookie](https://wicg.github.io/cookie-store/#set-a-cookie) with url, name, value, expires, domain, path, sameSite, and partitioned.

But the [set cookie](https://wicg.github.io/cookie-store/#set-a-cookie) steps say:
> If name’s [length](https://infra.spec.whatwg.org/#string-length) is 0 and value’s [length](https://infra.spec.whatwg.org/#string-length) is 0, then return failure.

That would mean `cookieStore.delete("")` should always rejects.

The [web platform tests](https://searchfox.org/mozilla-central/rev/27583d5afd9beea3a7551f35470cf188de1d368c/testing/web-platform/tests/cookie-store/change_eventhandler_for_no_name_and_no_value.https.window.js#20,38) currently assume that set throws for empty name and value, but delete with empty name works.

I think the spec needs to be updated to account for this corner case.

aamuley:
Perhaps we can update the spec to either not change the existing value of a cookie in the deletion algorithm or allow a UA to pick any non-empty value during the algorithm since it should not be persisted anyways? 

Any preference between those options?

DCtheTall:
The spec has been updated to instruct browsers to use a *non-empty* string where the value is implementation defined.

This is actually what Chromium's implementation was actually doing [for some time](https://crrev.com/c/2316622), I just never updated the spec after I landed that patch. Thanks @aamuley for taking care of that!

DCtheTall:
This seems to be fixed now. We have filed an issue to track only setting value for nameless cookies in Chromium due to the issue @aamuley pointed out about cookies whose name is 4KiB in size, https://crbug.com/427258226.

PR #254, opened 2025-05-01
The delete cookie algorithm should not overwrite a cookie with an empty value

aamuley:
Per https://github.com/WICG/cookie-store/issues/252 we should rework the spec so the delete cookie algorithm does not overwrite the cookie with an empty value. 

Otherwise, there is no way to delete a nameless cookie since the set cookie algorithm would fail for a nameless, valueless cookie.

***
<a href="https://pr-preview.s3.amazonaws.com/aamuley/cookie-store/pull/254.html" title="Last updated on May 5, 2025, 5:22 PM UTC (20dc427)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/254/37fb6f5...aamuley:20dc427.html" title="Last updated on May 5, 2025, 5:22 PM UTC (20dc427)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1210,7 +1210,7 @@ run the following steps:
     Note: The exact value of |expires| is not important for the purposes of this algorithm,
     as long as it is in the past.
 
-1. Let |value| be the empty string.
+1. Let |value| be any non-empty user-agent supplied string.
 1. Let |sameSite| be "{{CookieSameSite/strict}}".
 
     Note: The values for |value|, and |sameSite| will not be persisted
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1210,7 +1210,7 @@ run the following steps:
     Note: The exact value of |expires| is not important for the purposes of this algorithm,
     as long as it is in the past.
 
-1. Let |value| be the empty string.
+1. Let |value| be any non-empty user-agent supplied string.
```
annevk, line 1213: You want "implementation-defined" from Infra.

But also it seems fine to use "`dummy`" here and point out in a note that it's not observable.

aamuley: changed to use `implementation-defined`, there is a callout a couple lines down that seems sufficient 
`Note: The values for |value|, and |sameSite| will not be persisted by this algorithm.`
 but I can rework that if additional clarity would help.

annevk (approved):

Issue #245, opened 2025-01-23
Domain needs to be eagerly validated


annevk:
Colleagues discovered that Chromium rejects when attempting to set a cookie for a public suffix domain, rather than letting "receives a cookie" take care of that (which would result in resolving).

We will match that behavior as it's better for web developers.

But this needs to be defined.

cc @RupinMittal @bakulf

PR #256, opened 2025-05-05
Set cookie should parse domain into consistent format before setting

masnoble:
As discussed in #250, we should run the `domain` option through https://url.spec.whatwg.org/#concept-host-parser to ensure the domain is in a consistent and expected shape.

Will be adding WPTs to ensure there is coverage for various host types: https://chromium-review.googlesource.com/c/chromium/src/+/6613211

***
<a href="https://pr-preview.s3.amazonaws.com/masnoble/cookie-store/pull/256.html" title="Last updated on Jun 11, 2025, 6:31 PM UTC (bf376a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/256/9570d80...masnoble:bf376a3.html" title="Last updated on Jun 11, 2025, 6:31 PM UTC (bf376a3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1155,7 +1155,7 @@ run the following steps:
     1. If |name|, [=byte-lowercased=], [=byte sequence/starts with=] \``__host-`\`, then return failure.
     1. If |domain| [=is a registrable domain suffix of or is equal to|is not a registrable domain suffix of and is not equal to=] |host|, then return failure.
     1. Let |parsedDomain| be the result of [=host parser|host parsing=] |domain|.
-    1. Assert: |parsedDomain| is not a failure.
+    1. Assert: |parsedDomain| is not failure.
     1. Let |encodedDomain| be the result of [=UTF-8 encode|UTF-8 encoding=] |parsedDomain|.
     1. If the [=byte sequence=] [=byte sequence/length=] of |encodedDomain| is greater than the [=cookie/maximum attribute value size=], then return failure.
     1. [=list/Append=] \``Domain`\`/|encodedDomain| to |attributes|.
--- a/index.bs
+++ b/index.bs
@@ -710,16 +710,14 @@ The <dfn method for=CookieStore>set(|options|)</dfn> method steps are:
 1. Let |origin| be |settings|'s [=environment settings object/origin=].
 1. If |origin| is an [=opaque origin=], then return [=a promise rejected with=] a "{{SecurityError}}" {{DOMException}}.
 1. Let |url| be |settings|'s [=environment/creation URL=].
-1. Let |domain| be null.
-1. If |options|["{{CookieStoreSetOptions/domain}}"] is present, set |domain| to the result of [=host parser|host parsing=] |options|["{{CookieStoreSetOptions/domain}}"].
 1. Let |p| be [=a new promise=].
 1. Run the following steps [=in parallel=]:
     1. Let |r| be the result of running [=set a cookie=] with
-        |domain|,
         |url|,
         |options|["{{CookieInit/name}}"],
         |options|["{{CookieInit/value}}"],
         |options|["{{CookieInit/expires}}"],
+        |options|["{{CookieInit/domain}}"],
         |options|["{{CookieInit/path}}"],
         |options|["{{CookieInit/sameSite}}"], and
         |options|["{{CookieInit/partitioned}}"].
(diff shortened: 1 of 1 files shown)


aamuley:
could it make sense to do this in https://wicg.github.io/cookie-store/#set-cookie-algorithm instead of the [CookieStore.set(options)](https://wicg.github.io/cookie-store/#dom-cookiestore-set-options)? 
Step 9 in the set algo is where the actual comparison against the current domain happens so that might be a good place but maybe others have different opinions about this, @DCtheTall?

index.bs
```diff
@@ -1150,10 +1150,11 @@ run the following steps:
...
 1. If |domain| is not null, then run these steps:
     1. If |domain| starts with U+002E (`.`), then return failure.
-    1. If |host| does not equal |domain| and
-        |host| does not end with U+002E (`.`) followed by |domain|,
+    1. Let |parsedDomain| be the result of [=host parser|host parsing=] |domain|.
+    1. If |host| does not equal |parsedDomain| and
+        |host| does not end with U+002E (`.`) followed by |parsedDomain|,
         then return failure.
-    1. Let |encodedDomain| be the result of [=UTF-8 encode|UTF-8 encoding=] |domain|.
+    1. Let |encodedDomain| be the result of [=UTF-8 encode|UTF-8 encoding=] |parsedDomain|.
```
DCtheTall, line 1157: We should consider the case when host parsing returns a failure, since that is a possibility [according to the spec](https://url.spec.whatwg.org/#host-parsing).

DCtheTall (changes requested):

index.bs
```diff
@@ -1150,10 +1150,11 @@ run the following steps:
 1. Let |attributes| be a new [=/list=].
 1. If |domain| is not null, then run these steps:
     1. If |domain| starts with U+002E (`.`), then return failure.
-    1. If |host| does not equal |domain| and
-        |host| does not end with U+002E (`.`) followed by |domain|,
+    1. Let |parsedDomain| be the result of [=host parser|host parsing=] |domain|.
```
DCtheTall, line 1153: We should consider the case when host parsing returns a failure, since that is a possibility [according to the spec](https://url.spec.whatwg.org/#host-parsing).

index.bs
DCtheTall, line 1157: Should we maybe use the same encoding to store the host? I worry that this comparison will be incorrect in other parts of the spec now.

DCtheTall (changes requested):

index.bs
DCtheTall: Actually since `parsedDomain` is the input, we should be fine.

index.bs
masnoble: I added another line in the latest commit. PTAL!

index.bs
```diff
@@ -1150,10 +1150,12 @@ run the following steps:
...
+    1. If |parsedDomain| is a failure, then return failure.
+    1. If |host| does not equal |parsedDomain| and
```
annevk, line 1155: For equal you want https://url.spec.whatwg.org/#concept-host-equals

index.bs
```diff
@@ -1150,10 +1150,12 @@ run the following steps:
...
+        |host| does not end with U+002E (`.`) followed by |parsedDomain|,
```
annevk, line 1156: Use https://infra.spec.whatwg.org/#string-ends-with here?

annevk (commented):
This looks okay, but please identify the test coverage we have for this. This impacts how IP addresses are handled, uppercase hosts, IDNA hosts, percent-encoded hosts, ...

aamuley:
This domain parsing section seems to be getting a bit long and similar to https://html.spec.whatwg.org/#is-a-registrable-domain-suffix-of-or-is-equal-to. Could we just call that algo instead and return failure if it fails since it includes host-parsing the domain, handling the PSL which would fix https://github.com/WICG/cookie-store/issues/245, and some of the U+002E (.) parsing. Does that seem okay @annevk?

annevk:
If we can reuse that, that would be greatly preferable. Thanks for noticing!

index.bs
masnoble: Thanks!

now n/a since we call into the other algo where this check is done

index.bs
masnoble: Thanks!

now n/a since we call into the other algo where this check is done

annevk:
You will need to parse the input in the end, just like HTML does. Otherwise it's not canonicalized. (This makes me wonder if we want to slightly adjust HTML's algorithm so it returns the parsed input or failure...)

aamuley:
Spec changes seem good now!

index.bs
```diff
@@ -1153,10 +1153,10 @@ run the following steps:
 1. If |domain| is not null, then run these steps:
     1. If |domain| starts with U+002E (.), then return failure.
     1. If |name|, [=byte-lowercased=], [=byte sequence/starts with=] \``__host-`\`, then return failure.
-    1. If |host| does not equal |domain| and
-        |host| does not end with U+002E (.) followed by |domain|,
-        then return failure.
-    1. Let |encodedDomain| be the result of [=UTF-8 encode|UTF-8 encoding=] |domain|.
+    1. If |domain| [=is a registrable domain suffix of or is equal to|is not a registrable domain suffix of and is not equal to=] |host|, then return failure.
+    1. Let |parsedDomain| be the result of [=host parser|host parsing=] |domain|.
+    1. If |parsedDomain| is a failure, then return failure.
```
annevk, line 1158: Can we actually hit this step? Wouldn't the first algorithm have returned failure for this case? I think this should be an assert instead.

DCtheTall (approved):
LGTM

> This would work well for HTML and this algorithm at least. Parsing twice is also bit of a smell.

My 2 cents here is it's fine for the spec to call this twice and it can be up to implementers to compute the info needed for both of these steps with the same subroutine.

PR #259, opened 2025-05-20
Specify "__Host-" and "__Secure-" prefix failures

aamuley:
Per https://github.com/WICG/cookie-store/issues/248 add the cases where the CookieStore set algorithm would not resolve because of the prefix requirements to the spec. This is tested in [the set arguments wpt](https://github.com/web-platform-tests/wpt/blob/df1e48b497cc5dd19ba825ff273c5462b7962c87/cookie-store/cookieStore_set_arguments.https.any.js#L363) and [the special names wpt](https://github.com/web-platform-tests/wpt/blob/c9360813788a6f9d5ace7b00b18d170fc66f9d15/cookie-store/cookieStore_special_names.https.any.js)

Should this pr additionally change any language in https://wicg.github.io/cookie-store/#prefixes?

***
<a href="https://pr-preview.s3.amazonaws.com/aamuley/cookie-store/pull/259.html" title="Last updated on May 21, 2025, 5:24 PM UTC (12d9ea9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/259/abae748...aamuley:12d9ea9.html" title="Last updated on May 21, 2025, 5:24 PM UTC (12d9ea9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1157,11 +1157,12 @@ run the following steps:
     1. If the [=byte sequence=] [=byte sequence/length=] of |encodedDomain| is greater than the [=cookie/maximum attribute value size=], then return failure.
     1. [=list/Append=] \``Domain`\`/|encodedDomain| to |attributes|.
 1. If |expires| is given, then [=list/append=] \``Expires`\`/|expires| ([=date serialized=]) to |attributes|.
-1. If |path| does not start with U+002F (`/`), then return failure.
-1. If |path| does not end with U+002F (`/`), then append U+002F (`/`) to |path|.
-1. Let |encodedPath| be the result of [=UTF-8 encode|UTF-8 encoding=] |path|.
-1. If the [=byte sequence=] [=byte sequence/length=] of |encodedPath| is greater than the [=cookie/maximum attribute value size=], then return failure.
-1. [=list/Append=] \``Path`\`/|encodedPath| to |attributes|.
+1. If |path| is not null, then run these steps:
+    1. If |path| does not start with U+002F (`/`), then return failure.
+    1. Let |encodedPath| be the result of [=UTF-8 encode|UTF-8 encoding=] |path|.
+    1. If the [=byte sequence=] [=byte sequence/length=] of |encodedPath| is greater than the [=cookie/maximum attribute value size=], then return failure.
+    1. [=list/Append=] \``Path`\`/|encodedPath| to |attributes|.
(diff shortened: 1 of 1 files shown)


annevk (commented):
I think the non-normative language is still okay. Though I'm not sure "what more widely adopted" is doing there as nobody has plans to change `document.cookie` that I know of. But that should be addressed separately.

index.bs
```diff
@@ -1141,15 +1141,18 @@ run the following steps:
 
     ISSUE(httpwg/http-extensions#1593): Note that it's up for discussion whether these character restrictions should also apply to |expires|, |domain|, |path|, and |sameSite| as well.
 
-1. If |name|'s [=string/length=] is 0 and |value| contains U+003D (`=`), then return failure.
-1. If |name|'s [=string/length=] is 0 and |value|'s [=string/length=] is 0, then return failure.
+1. If |name|'s [=string/length=] is 0:
+    1. If |value| contains U+003D (`=`), then return failure.
+    1. If |value|'s [=string/length=] is 0, then return failure.
+    1. If |value|, [=byte-lowercased=], [=byte sequence/starts with=] `__host-` or `__secure-`, then return failure.
```
annevk, line 1147: The proper syntax for this would be (otherwise it's not clear they are Infra byte sequences):
```suggestion
    1. If |value|, [=byte-lowercased=], [=byte sequence/starts with=] \``__host-`\` or \``__secure-`\`, then return failure. 
```
This also applies to the changes below.

index.bs
```diff
@@ -1141,15 +1141,18 @@ run the following steps:
 
     ISSUE(httpwg/http-extensions#1593): Note that it's up for discussion whether these character restrictions should also apply to |expires|, |domain|, |path|, and |sameSite| as well.
 
-1. If |name|'s [=string/length=] is 0 and |value| contains U+003D (`=`), then return failure.
-1. If |name|'s [=string/length=] is 0 and |value|'s [=string/length=] is 0, then return failure.
+1. If |name|'s [=string/length=] is 0:
+    1. If |value| contains U+003D (`=`), then return failure.
```
annevk, line 1145: Existing issue that maybe should be fixed separately throughout, Infra doesn't use `<code>` for rendering of standalone code points:
```suggestion
    1. If |value| contains U+003D (=), then return failure.
```

annevk (approved):
This looks okay from my perspective, modulo some syntax nits.

DCtheTall (approved):

PR #261, opened 2025-05-30
Only use implementation defined value for deleting nameless cookies

aamuley:
This accounts for the case where the name is the maximum size and a cookie's value must be 0B.  https://github.com/WICG/cookie-store/issues/252 

WPT for this will be added in https://chromium-review.googlesource.com/c/chromium/src/+/6604039

***
<a href="https://pr-preview.s3.amazonaws.com/aamuley/cookie-store/pull/261.html" title="Last updated on May 30, 2025, 4:39 PM UTC (6a2b640)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/cookie-store/261/abae748...aamuley:6a2b640.html" title="Last updated on May 30, 2025, 4:39 PM UTC (6a2b640)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1209,7 +1209,9 @@ run the following steps:
     as long as it is in the past.
 
 1. Let |value| be the empty string. 
+
 1. If |name|'s [=string/length=] is 0, then set |value| to any non-empty [=implementation-defined=] string.
+
 1. Let |sameSite| be "{{CookieSameSite/strict}}".
 
     Note: The values for |value|, and |sameSite| will not be persisted
--- a/index.bs
+++ b/index.bs
@@ -1157,11 +1157,12 @@ run the following steps:
     1. If the [=byte sequence=] [=byte sequence/length=] of |encodedDomain| is greater than the [=cookie/maximum attribute value size=], then return failure.
     1. [=list/Append=] \``Domain`\`/|encodedDomain| to |attributes|.
 1. If |expires| is given, then [=list/append=] \``Expires`\`/|expires| ([=date serialized=]) to |attributes|.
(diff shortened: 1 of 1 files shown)


DCtheTall (approved):
LGTM and nice catch of this subtle bug

index.bs
```diff
@@ -1208,7 +1208,8 @@ run the following steps:
     Note: The exact value of |expires| is not important for the purposes of this algorithm,
     as long as it is in the past.
 
-1. Let |value| be any non-empty [=implementation-defined=] string.
+1. If |name|'s [=string/length=] is 0, let |value| be any non-empty [=implementation-defined=] string.
+1. Otherwise, let |value| be the empty string.
```
annevk, line 1212: Usually this is the preferred style:
```suggestion
1. Let |value| be the empty string. 
1. If |name|'s [=string/length=] is 0, then set |value| to any non-empty [=implementation-defined=] string.
```
Or alternatively:
```suggestion
Let |value| be the empty string if |name| is not the empty string; otherwise any non-empty [=implementation-defined=] string.
```

index.bs
```diff
@@ -1208,7 +1208,8 @@ run the following steps:
     Note: The exact value of |expires| is not important for the purposes of this algorithm,
     as long as it is in the past.
 
-1. Let |value| be any non-empty [=implementation-defined=] string.
+1. Let |value| be the empty string. 
+1. If |name|'s [=string/length=] is 0, then set |value| to any non-empty [=implementation-defined=] string.
```
annevk, line 1212: Final nit: you need to add newlines between the steps to match the existing style of the document.

annevk (approved):

PR #269, opened 2025-07-24
Strip whitespace from cookie name and value when evaluating prefixes

yoavweiss:
As pointed out by @annevk at https://github.com/WebKit/WebKit/pull/48318, we should verify that the prefix validation can deal with whitespaces. This PR handles that.

***
<a href="https://whatpr.org/cookiestore/269.html" title="Last updated on Jul 29, 2025, 6:01 AM UTC (c877ed6)">Preview</a> | <a href="https://whatpr.org/cookiestore/269/0fc20e4...c877ed6.html" title="Last updated on Jul 29, 2025, 6:01 AM UTC (c877ed6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1137,6 +1137,8 @@ optional |expires|,
 |partitioned|
 run the following steps:
 
+1. Set |name| to |name|, [=stripped leading and trailing ASCII whitespace=].
+1. Set |value| to |value|, [=stripped leading and trailing ASCII whitespace=].
 1. If |name| or |value| contain U+003B (;), any [=C0 control=] character except U+0009 TAB, or U+007F DELETE, then return failure.
 
     ISSUE(httpwg/http-extensions#1593): Note that it's up for discussion whether these character restrictions should also apply to |expires|, |domain|, |path|, and |sameSite| as well.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1137,6 +1137,8 @@ optional |expires|,
 |partitioned|
 run the following steps:
 
+1. Set |name| to |name|, [=stripped leading and trailing ASCII whitespace=].
+1. Set |value| to |value|, [=stripped leading and trailing ASCII whitespace=].
```
annevk: This is wrong. Because you now you strip newlines and `\f` for which the next step would have returned failure. And we want to continue to return failure for those. You either have to do this after the next step or only strip U+0009 and U+0020.

yoavweiss: Fixed to only strip 0x9 and 0x20

aamuley:
Hey, are we also planning on changing https://wicg.github.io/cookie-store/#set-cookie-algorithm or excluding the tab character from this PR? 

> If name or value contain U+003B (;), any [C0 control](https://infra.spec.whatwg.org/#c0-control) character except U+0009 TAB, or U+007F DELETE, then return failure.

and also could this PR use `U+0009 TAB` rather than `U+0009` for clarity?

annevk:
@DCtheTall should probably do the final review. Does this have test coverage too now?

yoavweiss:
> Does this have test coverage too now

Not yet, but soon

annevk (approved):

PR #302, opened 2026-07-21
Set navigation conditions potentially changed when setting cookie

llannasatoll:
This PR updates the spec to integrate with the new `navigation conditions potentially changed` concept introduced in https://github.com/whatwg/html/pull/11765.

- [x] At least two implementers are interested (and none opposed):
   * Chromium (https://github.com/whatwg/html/issues/11743)
   * …
- [x] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://chromium-review.googlesource.com/c/chromium/src/+/5856811
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chromium: https://crbug.com/366060351
   * Gecko: …
   * WebKit: …
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …
- [ ] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

(See [WHATWG Working Mode: Changes](https://whatwg.org/working-mode#changes) for more details.)

***
<a href="https://whatpr.org/cookiestore/302.html" title="Last updated on Aug 4, 2026, 4:33 AM UTC (9b8ca08)">Preview</a> | <a href="https://whatpr.org/cookiestore/302/54559c0...9b8ca08.html" title="Last updated on Aug 4, 2026, 4:33 AM UTC (9b8ca08)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -29,6 +29,8 @@ Level: 1
 spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
     type: dfn; text: time values; url: sec-time-values-and-time-range
     type: dfn; text: ToString; url: sec-tostring
+spec: html; urlPrefix: https://html.spec.whatwg.org/
+    type: dfn; text: navigation conditions potentially changed; url: #navigation-conditions-potentially-changed
 </pre>
 
 <pre class=link-defaults>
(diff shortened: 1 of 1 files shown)


DCtheTall:
This feature seems to apply to all cookies, not just the Cookie Store API. It probably makes sense to update the spec in one place where the cookie storage model is invoked.

@annevk I think this type of thing belongs in Fetch when we move more of the cookie spec into that. The flag in this feature would be a side effect of Cookie Store API calling methods in the Fetch spec, since I do not see a reason this API needs to be aware of this flag. WDYT?

annevk:
I think it's just cookie APIs (so this and `document.cookie`) that can cause this flag to be set. Not anything network-based. If that's correct having it here might be okay, though perhaps we should still have an abstraction that all APIs can use that takes care of this.

